PR #236, opened 2022-03-25
Jest mock onLayout

fortmarek:
## Description

Testing `FlashList` currently is not ideal. Users tend to test the items in the list but `FlashList` does not render anything until `onLayout` is called from the underlying scroll component (or when [`estimatedListSize`](usage#estimatedlistsize) is defined). It leads to boilerplate code like [this](https://github.com/Shopify/mobile/pull/4644/files#diff-036f0c7639980915233114398aba2f02f844fcb44b3ecc44c32bb5dd829f40b5R81-R91).

We should make testing the `FlashList` component easier by triggering the `onLayout` event ourselves. With this PR, users can include `require("@shopify/flash-list/jestSetup");` in their `setup-jest.js` file and then test the component the same way they would have `FlatList`.

The drawback of the current solution is that the list size is hardcoded. We can either make it configurable by:
- providing global values in the new `setupJest.js` file
- making it easier to trigger a new `onLayout` event

I prefer the latter solution but doing so will be easier once we implement [`getNativeScrollRef`](https://reactnative.dev/docs/flatlist#getnativescrollref) method. For now, providing a sensible default is good enough IMHO.

I have tried creating a test case in the fixture app but unfortunately, it turns out to be not that simple since we link the library instead of using it as an external dependency.

I have [prepared a branch](https://github.com/Shopify/flash-list/tree/jest-mock-test-fixture) that is the same as the one for this PR with the exception of [this](https://github.com/Shopify/flash-list/commit/8432809b5d75f1780a4792a89779ec1c6f204c8c) commit.

To test out that the mock works, you can run:
```
yarn
cp ../jestSetup.js node_modules/@shopify/flash-list/
yarn jest
// comment out the single line in `setup.js` file
yarn jest // tests should fail because we are not autotriggerin onLayout event
```

I had two issues with using the `@shopify/flash-list` library as a link.

When I used the `setupJest.js` that is in the root, I was getting the following error:
```
 FAIL  src/__tests__/Twitter.test.tsx
  ● Test suite failed to run

    Cannot find module '@shopify/flash-list' from '../jestSetup.js'

    Require stack:
      /Users/marekfort/src/github.com/Shopify/flash-list/jestSetup.js
      setup.js

    > 1 | jest.mock("@shopify/flash-list", () => {
        |                                                                                 ^
      2 |   const ActualFlashList = jest.requireActual("@shopify/flash-list").FlashList;
      3 |   class MockFlashList extends ActualFlashList {
      4 |     componentDidMount() {

      at Resolver.resolveModule (node_modules/jest-resolve/build/resolver.js:324:11)
      at Object.<anonymous> (../jestSetup.js:1:97)
```

But even when I moved `setupJest.js` file contents to `fixture/setup.js`, I still got the following error:
```
 FAIL  src/__tests__/Twitter.test.tsx
  ● Test suite failed to run

    Invariant Violation: __fbBatchedBridgeConfig is not set, cannot invoke native modules

      at invariant (../node_modules/invariant/invariant.js:40:15)
      at Object.<anonymous> (../node_modules/react-native/Libraries/BatchedBridge/NativeModules.js:181:3)
      at Object.<anonymous> (../node_modules/react-native/Libraries/TurboModule/TurboModuleRegistry.js:11:23)
```

I don't want to spend more time with setting up the test in the fixture but if you have any ideas how to resolve any of those issues above, let me know.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-- Added a new for easier testing of components with `FlashList`
+- Added a new mock for easier testing of components with `FlashList`
   - https://github.com/Shopify/flash-list/pull/236
 
 ## [0.4.0] - 2022-03-23


tycobbb:
any progress on an easy way to trigger a different rect in tests? the fixed height in the mock makes testing the rendered contents of the scroll view with a large `initialScrollIndex` pretty challenging.

Issue #196, opened 2022-03-18
Explore Fabric support


naqvitalha:
# What
With Fabric now being available for use we should look at making FlashList compatible with it.

Reference link: https://reactnative.dev/blog/2022/03/15/an-update-on-the-new-architecture-rollout

naqvitalha:
v1 and v2 both support fabric.

PR #340, opened 2022-04-25
Fabric iOS support

fortmarek:
## Description

Related: https://github.com/Shopify/flash-list/issues/196

This is an initial draft for Fabric support. I have hit a couple of issues we'll need to resolve, some of them are not a flash-list issue specifically:
- `@react-navigation/stack` seems not to be compatible with Fabric (I get the same error as [here](https://github.com/facebook/react-native/issues/33668))
  - for now, I moved to `@react-navigation/native-stack` which seems to work and is not an issue (fixture-specific only)
- the hierarchy of views have changed, we'll need to change accessing `AutoLayoutView` subviews
- `FlashListPerformanceView` is not migrated to Fabric 
- `onBlankAreaEvent` is currently not called in Fabric
- migrate `reanimated` to v3
- only elements during the first render are loaded - when scrolling, new ones are not added. This is possibly issue in `recyclerlistview` (possibly some assumption is no longer valid on the new arch)
- it is not possible to touch elements wrapped inside a Fabric native view (reproduced in the RN tester app [here](https://github.com/facebook/react-native/compare/main...fortmarek:my-wrapper-view))

Minor issues (don't need to be resolved):
- Switching between Fabric and non-Fabric is confusing. You will get build errors because of the code generated by Fabric which might not be compatible with your pre-Fabric app.

Resolved issues:
- Some dependencies (listed when running `USE_FABRIC=1 RCT_NEW_ARCH_ENABLED=1 bundle exec pod install`) in [react_native_pods](https://github.com/facebook/react-native/blob/main/scripts/react_native_pods.rb) turned on Fabric must have `modular_headers: true` when using Swift code
  - If you add that setting for relevant dependencies, the modular headers for `React-RCTFabric` and `React-Core` conflict
  - both of these will need to be fixed in `react-native` directly (it seems Swift support has not been properly tested)
  - both fixed [here](https://github.com/facebook/react-native/pull/33743)
- `reanimated` does [not support](https://github.com/software-mansion/react-native-reanimated/issues/2991#issuecomment-1035637363) Fabric which causes issues for our fixture app (Reminders example). It should be ready along with 0.69 release of `react-native` *Update*: It does now 🎉 

I will continue working on those we can tackle and create appropriate issues or PRs (if possible) for those that are outside are repos.

### Resources used during the migration:
- https://reactnative.dev/docs/next/new-architecture-library-ios (and other parts of the new architecture documentation)
- https://react-native-community.github.io/upgrade-helper/?from=0.67.4&to=0.68.1
- [Commit](https://github.com/software-mansion/react-native-gesture-handler/commit/0d1a917a3de5281656276e05bcbff1d832d8d624) bringing support for Fabric inside `react-native-gesture-handler`. The great thing here is that the library support both Fabric and Paper whereas the official documentation assumes you will want to support Fabric only.
- [Example app](https://github.com/react-native-community/RNNewArchitectureApp) showcasing migrating to Fabric along with a custom Fabric native component
- https://github.com/react-native-svg/react-native-svg/pull/1754

- [ ]

## Checklist


fortmarek:
Closing in favor of https://github.com/Shopify/flash-list/pull/549

PR #381, opened 2022-05-05
Implement missing scroll component methods

fortmarek:
## Description

Implementing missing `getNativeScrollRef` and `getScrollResponder`. Both methods currently rely on a private property from `recyclerlistview`.

I was not able to find out whether it's possible to access `_scrollComponent?._scrollViewRef` via some public method. We have the following ways forward:
- find a way how to use existing public API of `recyclerlistview` to get the `_scrollViewRef`
- make those methods public in `recyclerlistview` (see [here](https://github.com/facebook/react-native/blob/c231d5e371cbddf0e54bf84f46336b22bbdb5367/Libraries/Lists/VirtualizedList.js#L564-L586) the implementation in `VirtualizedList`)
- keep it the way it is and access private `recyclerlistiview` properties (not a good idea for long term)

@naqvitalha do you have any suggestions regarding this? 🤞 

1. Check out this branch
2. Apply [this](https://github.com/Shopify/flash-list/commit/d742bdd1964d55b9fe1c58a71a57094557cd8c5d) commit
3. Go to Twitter example for both `FlashList` and `FlatList`
4. You should see logs for all the methods 👀 

## Checklist

### DIFF
--- a/documentation/docs/fundamentals/usage.md
+++ b/documentation/docs/fundamentals/usage.md
@@ -484,7 +484,7 @@ Provides a reference to the underlying scroll component
 
 ### `getScrollResponder()`
 
-```jsx
+```ts
 getScrollResponder(): JSX.Element | null | undefined;
 ```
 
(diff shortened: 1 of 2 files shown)


fobos531:
Hey @fortmarek @naqvitalha , what is the verdict on this PR? would love to see the proposed methods get merged.

PR #444, opened 2022-06-09
Web support (#1)

### LINKED ISSUE #23: Implement FlashList for web
We will need to build the following to make it work on web

- AutoLayoutView for web
- Cell container for web
- Blank events (optional)

In the meantime we can replace them with views and it should still work

naqvitalha:
## Description

resolves #23 

Note: There are actually very few changes. Most of the files have just been moved and there's boilerplate for web sample.

- I've replaced AutoLayout and CellContainer with View and div respectively. This means that there won't be layout corrections, instead, we'll have shift animations. There also won't be blank area events.
- Fixed inverted mode on web. Our transform for inverted mode was overriding transform applied by RLV.
- Added a sample for web

- [ ] Goto `fixture/web-app` and run `yarn web`

https://user-images.githubusercontent.com/7811728/172900517-09081e03-c147-4511-b79d-17024882441f.mov

## Checklist

### DIFF
--- a/fixture/web-app/List.tsx
+++ b/fixture/web-app/List.tsx
@@ -2,7 +2,7 @@
 Sample list for web
  */
 import React, { useEffect, useRef, useState } from "react";
-import { View, Text, Pressable } from "react-native";
+import { View, Text, Pressable, StyleSheet } from "react-native";
 import { FlashList, ListRenderItemInfo } from "@shopify/flash-list";
 
 const ListItem = ({ index }: { index: string }) => {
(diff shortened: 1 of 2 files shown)


noskovvkirill:
> Hi @naqvitalha, I'm getting the error `Cannot use import statement outside a module` on react-native-web. Not sure what is going on there.

Had the same issue earlier today. It was solved by disabling SSR and dynamically importing it (I'm using Next.js). Not sure if it's applicable in your case, but might be.

PR #454, opened 2022-06-14
Added render target info to renderItem callback (#4)

### LINKED ISSUE #426: Detect dummy element mount
### Description
To measure horizontal list height we mount an extra item but we don't provide a way to developers to detect this. This can lead to issues if developers implement any kind of tracking.
We should provide an extra field using which this behaviour can be detected.

naqvitalha:
## Description

Review after #453 has been merged and this branch has been updated. Otherwise, you will see too many files changed.

resolves #426 

When we implemented horizontal list measurement we introduced a flow where an extra item is mounted outside the list. While visually nothing weird happens, most apps implement some sort of tracking on component mount or update. There is no way to handle such cases right now. Apps with analytics will end up firing extra events.

In this PR I've introduced three render targets:

- `Cell`: This is the list item
- `StickyHeader`: Can also be used to style the sticky header in a different way
- `Measurement`: Can be ignored in analytics

N/A

## Checklist


ElviraBurchik (approved):
Makes sense to me. Let's merge it :)

PR #481, opened 2022-06-30
Add kotlin-gradle-plugin to buildscript in project build.gradle

brentvatne:
## Description

Fixes the error many developers will encounter when using this library in their project, which is explained in the README:

> If you get the error `Plugin with id 'kotlin-android' not found` while building on Android, go to `android/build.gradle` and add `classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.0")` inside your `dependencies` block. Please change the plugin version as per your needs.

This PR also removes the above paragraph from the README as it's no longer needed.

It does this by adding the dependency in the buildscript in the project build.gradle. We use a similar approach in Expo modules, for example in [expo-gl](https://github.com/expo/expo/blob/98e3d1132335b3a16bdd954103618e9f06ec2e49/packages/expo-gl/android/build.gradle#L8-L36), and you will find [the same pattern in other libraries too](https://github.com/mrousavy/react-native-blurhash/blob/d72de0e175783d5e540e1e2e689c9584159039a0/android/build.gradle#L1-L15).

I don't think this applies to tophatting as there is no change to runtime behavior. Here's how you can reproduce the issue and test the fix:

```
npx react-native init myapp
cd myapp
yarn add @shopify/flash-list
yarn android # notice the build will fail with the error mentioned above
# now copy the diff from this pr over to node_modules/@shopify/flash-list/android/build.gradle
yarn android # notice the build will succeed
```

image

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Add kotlin-gradle-plugin to buildscript in project build.gradle
+  - https://github.com/Shopify/flash-list/pull/481
+
 ## [1.0.2] - 2022-06-30
 
 - Minor changes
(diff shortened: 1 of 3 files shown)


naqvitalha (approved):
LGTM

fortmarek:
Hey @brentvatne 👋 

Thanks a lot for the PR, I've checked it locally and it does, indeed, fix the issue ❤️

PR #487, opened 2022-07-01
Fix missing return for `getItemType` in website snippet example

jamsch:
## Description

Function signature shows that it can return a string, number, or undefined. It seems like the intention in this snippet is to return `item.type` so that it can be properly recycled.

N/A

N/A

## Checklist

(Not applicable?)


fortmarek (approved):
Good catch @jamsch, you are right, there should be a `return` in order for the prop to work correctly 🙂

Issue #495, opened 2022-07-01
Build fails on Android with error "No signature of method: org.gradle.internal.extensibility.DefaultExtraPropertiesExtension.kotlinVersion() is applicable for argument types: () values: []"


micheleb:
Hi there 👋🏻 and thanks for the great library!

I'm getting the error reported in the title when building my RN app on Android, which is due to line 18 in flash-list's `build.gradle`:

```
    ext._kotlinVersion = rootProject.ext.has("kotlinVersion") ?  rootProject.ext.kotlinVersion() : '1.5.30'
```

I already had a `kotlinVersion` variable defined in my project, like this:

```
buildscript {
    ext {
        // ... many other variables, and then
        kotlinVersion = "1.6.10"
```

so that `rootProject.ext.kotlinVersion()` is invoking my string variable as a function, which is causing the issue.

Removing the `kotlinVersion` variable declaration from my `build.gradle` fixes the issue, but I was wondering if it's more common to have that variable defined as a function? I thought my setup was pretty standard there 🤷🏻 

Thanks!

AndreiCalazans:
@brentvatne it seems like [your recent change](https://github.com/Shopify/flash-list/pull/481) is causing this issue for those of us that set the kotlinVersion. Question, is there a better way to set the kotlinVersion to make it compatible with your change? 

```
buildscript {
    ext {
...
        kotlinVersion = '1.6.10'
   ... 
} 
```

AndreiCalazans:
[...]

PR #496, opened 2022-07-01
fix get kotlinVersion

AndreiCalazans:
## Description

Fixes https://github.com/Shopify/flash-list/issues/495. See issue it fully describes the problem how this solves it. 

```
npx react-native init myapp
cd myapp
yarn add @shopify/flash-list
yarn android # notice the build will fail with the error mentioned above
# now copy the diff from this pr over to node_modules/@shopify/flash-list/android/build.gradle
yarn android # notice the build will succeed
```
## Checklist


brentvatne:
🤦

naqvitalha:
This is already fixed in v1.0.4

PR #502, opened 2022-07-03
Allow providing an external scrollview.

gorhom:
## Description

Fixes (issue #)

This PR will allow passing an external scrollview to the `recyclerlistview`, which is needed to allow gesture interactions with `react-native-gesture-handler` especially on **Android**.

- Prop description and docs.

#### Before 

https://user-images.githubusercontent.com/4061838/177045508-f5b21abc-7604-4292-9835-c8c26018afd4.mov

#### After 

https://user-images.githubusercontent.com/4061838/177045515-b311fcbd-c06b-43b0-9254-b0ca9be93620.mov
 
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## [1.0.4] - 2022-07-02
 
 - Build fix for Android projects having `kotlinVersion` defined in `build.gradle`.
+- Allow providing an external scrollview.
 
 ## [1.0.3] - 2022-07-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 - Build fix for Android projects having `kotlinVersion` defined in `build.gradle`.
 - Allow providing an external scrollview.
+  - https://github.com/Shopify/flash-list/pull/502
 
 ## [1.0.3] - 2022-07-01
 
--- a/documentation/docs/fundamentals/usage.md
+++ b/documentation/docs/fundamentals/usage.md
@@ -96,12 +96,12 @@ estimatedItemSize?: number;
 
 ---
 
-### `ExternalScrollViewComponent`
+### `renderScrollComponent`
 
 Rendered as the main scrollview. Its contract for the scroll event should match the native scroll event contract, i.e. `scrollEvent = { nativeEvent: { contentOffset: { x: offset, y: offset } } }`
 
 ```ts
-return <FlashList ExternalScrollViewComponent={Animated.ScrollView} />;
+return <FlashList renderScrollComponent={Animated.ScrollView} />;
 ```
 
 ### `CellRendererComponent`
--- a/documentation/docs/fundamentals/usage.md
+++ b/documentation/docs/fundamentals/usage.md
@@ -96,6 +96,14 @@ estimatedItemSize?: number;
 
 ---
 
+### `ExternalScrollViewComponent`
+
+Rendered as the main scrollview. Its contract for the scroll event should match the native scroll event contract, i.e. `scrollEvent = { nativeEvent: { contentOffset: { x: offset, y: offset } } }`
+
+```ts
+return <FlashList ExternalScrollViewComponent={Animated.ScrollView} />;
+```
+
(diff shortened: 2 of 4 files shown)


nandorojo:
Thoughts on renaming the prop to `ScrollViewComponent`?

gorhom:
@naqvitalha sounds better! i have updated the pr 👍

src/FlashListProps.ts
```diff
@@ -122,6 +124,14 @@ export interface FlashListProps<TItem> extends ScrollViewProps {
    */
   ListHeaderComponentStyle?: StyleProp<ViewStyle> | undefined;
 
+  /**
+   * Rendered as the main scrollview. Its contract for the scroll event should match the native scroll event contract, i.e.
+   * `scrollEvent = { nativeEvent: { contentOffset: { x: offset, y: offset } } }`
+   */
+  renderScrollComponent?:
```
naqvitalha, line 131: Let's use default scroll view props from RN. That should be fine. All this complexity in RLV is due to its web only implementation. FlashList doesn't depend on it so using ScrollView props should be fine.

src/FlashList.tsx
```diff
@@ -350,6 +351,7 @@ class FlashList<T> extends React.PureComponent<
           windowCorrectionConfig={this.getUpdatedWindowCorrectionConfig()}
           itemAnimator={this.itemAnimator}
           suppressBoundedSizeException
+          externalScrollView={renderScrollComponent as any}
```
naqvitalha, line 354: Typecast to BaseScrollView instead

documentation/docs/fundamentals/usage.md
```diff
@@ -96,6 +96,14 @@ estimatedItemSize?: number;
 
 ---
 
+### `renderScrollComponent`
```
naqvitalha, line 99: nit: Props are sorted alphabetically. Moving this would be helpful.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## [1.0.4] - 2022-07-02
 
 - Build fix for Android projects having `kotlinVersion` defined in `build.gradle`.
+- Allow providing an external scrollview.
```
naqvitalha, line 13: Add link to this PR here

naqvitalha (commented):
Few more comment. Looks good otherwise.

src/FlashList.tsx
gorhom: i had to cast it to `RecyclerListViewProps["externalScrollView"]` instead

```ts
    externalScrollView?: {
        new (props: ScrollViewDefaultProps): BaseScrollView;
    };
```

gorhom:
@naqvitalha i have addressed you comments in the latest commit 👍

naqvitalha (approved):
Thanks for the PR. Looks great!

sdfricke:
> Using this prop this way: `renderScrollComponent={props => <ScrollView {...props} />}` I get this warning: `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

I'm seeing the same issue

hirbod:
```tsx
const ScrollViewWithRef = React.forwardRef((props, ref) => (
  <ScrollView {...props} ref={ref} />
)) as typeof ScrollView;
```

sdfricke:
> renderScrollComponent={props => <ScrollView {...props} />}

Thank you for the suggestion.  I'm not quite grasping how that would be implemented in the context of this simple example:
```renderScrollComponent={props => <ScrollView {...props} />}```

I've tried a couple different ways but getting quite a few errors so I'm guessing I'm not implementing that suggestion as intended.

sdfricke:
> ```tsx
> const ScrollViewWithRef = React.forwardRef((props, ref) => (
>   <ScrollView {...props} ref={ref} />
> )) as typeof ScrollView;
> ```

Thank you for the suggestion. I'm not quite grasping how that would be implemented in the context of this simple example:
renderScrollComponent={props => <ScrollView {...props} />}

I've tried a couple different ways but getting quite a few errors so I'm guessing I'm not implementing that suggestion as intended.

hirbod:
Please create a repo on Snack so I can have a look

sdfricke:
@hirbod it turns out I was on older version.  It looks like the issue I was seeing was fixed in version 1.4.0 as described here: https://github.com/Shopify/flash-list/issues/616

PR #504, opened 2022-07-03
fix typo

kku77:
Fix typo of FlashList


rish:
This isn't a typo. The sentence is actually referring to the standard library `FlatList` component if you read it carefully.

naqvitalha:
That's not a type. We're actually referring to FlatList in that statement.

PR #506, opened 2022-07-05
Handle dark theme

naqvitalha:
## Description

Prevents text in fixture from disappearing in light theme


fixture/android/app/src/main/res/values/styles.xml
```diff
@@ -1,8 +1,8 @@
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
 
     <!-- Base application theme. -->
-    <style name="AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
```
ElviraBurchik, line 4: I don't know how these lines on Android work - could you please explain to me what do they do?

naqvitalha: DayNight theme is supposed to support dark mode but given we explicitly set text colors etc it doesn't really work.

Switching to light theme explicitly so that app's text is visible in light mode. Today on Android 11 text is not visible (white on white)

The other line prevents Android from forcing dark mode by overriding some styles on its own. Since, this is fixture, it's not really required and we can just disable it to avoid surprises.

Issue #133, opened 2022-03-01
Add AutoLayout test cases on iOS


naqvitalha:
# What
We don't have tests for main layout correction algorithms on iOS. This issue is to track it's addition. The methods is call `clearGaps` inside `AutoLayoutView`

PR #519, opened 2022-07-08
Fix footer layout

naqvitalha:
## Description

Fixes (issue #)

- [ ]

## Checklist


src/__tests__/FlashList.test.tsx
```diff
@@ -606,7 +606,9 @@ describe("FlashList", () => {
 
     // items widths before layout manager change should be 400
     flashList.findAll(CellContainer).forEach((cell) => {
-      expect(cell.instance.props.style.width).toBe(400);
+      if (cell.props.index !== -1) {
```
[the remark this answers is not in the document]
naqvitalha: -1 is the footer wrapper. That's irrelevant for this test.

android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutView.kt
```diff
@@ -66,6 +70,72 @@ class AutoLayoutView(context: Context) : ReactViewGroup(context) {
...
+
+    private fun getParentScrollView(): View? {
+        var alParent = parent;
+        while (alParent != null) {
+            if (alParent is ScrollView || alParent is HorizontalScrollView) {
+                return alParent as View
+            }
+            alParent = alParent.parent;
+        }
+        return null
```
fortmarek, line 135:
```suggestion
        return generateSequence(parent) { it.parent }
                .firstOrNull { it is ScrollView || it is HorizontalScrollView } as? View
```

You could also rewrite it with this less-verbose version which would be more similar to what we do on iOS. It'd be nice to keep those two implementations as close as possible.

android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutView.kt
```diff
@@ -66,6 +70,72 @@ class AutoLayoutView(context: Context) : ReactViewGroup(context) {
...
+        return alShadow.lastMaxBoundOverall - autoLayoutEnd
+    }
+
+    private fun getFooter(): View? {
+        return (parent as ViewGroup)?.let {
+            val count = it.childCount;
+            for (i in 0 until count) {
+                val view = it.getChildAt(i)
+                if (view is CellContainer && view.index == -1) {
+                    return@let view as View
```
fortmarek, line 120:
```suggestion
                    return@let view
```

ios/Sources/AutoLayoutView.swift
```diff
@@ -215,4 +224,44 @@ import UIKit
...
+		} else {
+			footer.frame.origin.y += diff
+			frame.size.height += diff
+			superview?.frame.size.height += diff
+		}
+	}
+
+    func getFooterDiff() -> CGFloat {
+        if subviews.count == 0 {
+            lastMaxBoundOverall = 0
```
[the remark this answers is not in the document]
naqvitalha: clearGaps only runs if you have more than one child. I also wanted to keep footer stuff disconnected and not refactor too much. In case of any issue we just have to stop calling fixFooter()
Lack of unit tests isn't helping :(

ios/Sources/AutoLayoutView.swift
```diff
@@ -39,16 +39,19 @@ import UIKit
...
+    /// Tracks where the last pixel is drawn in the overall
+    private var lastMaxBoundOverall: CGFloat = 0
     /// Tracks where the last pixel is drawn in the visible window
     private var lastMaxBound: CGFloat = 0
     /// Tracks where first pixel is drawn in the visible window
     private var lastMinBound: CGFloat = 0
 
     override func layoutSubviews() {
         fixLayout()
+        fixFooter()
```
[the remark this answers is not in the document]
naqvitalha: We need to pick #133 there are no mocks to build upon right now. I moved on to other low-hanging wins for the same reason.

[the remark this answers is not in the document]
fortmarek: ouch, you are right 😞 let's skip for now then.

android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutView.kt
naqvitalha: Good suggestion. I'm skipping for now just to understand these sequences better. Nothing beats while when it comes to speed.

fixture/src/contacts/data/contacts.ts
```diff
@@ -411,8 +411,8 @@ const contacts = contactsData.reduce((contactsMap, contact) => {
   return contactsMap;
 }, new Map<string, Contact[]>());
 contacts.forEach(
-  (contacts: Contact[], key: string, map: Map<string, Contact[]>) => {
-    const sortedContacts = contacts.sort((aContact, bContact) =>
+  (_contacts: Contact[], key: string, map: Map<string, Contact[]>) => {
+    const sortedContacts = _contacts.sort((aContact, bContact) =>
```
fortmarek, line 415: Not sure if this change was necessary since we don't want to access the outer `contacts` but I'm not against keeping it this way as well.

android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutView.kt
fortmarek: I dug a little deeper to see the Kotlin Bytecode, just out of interest:

This is bytecode for the version with `while`:
```
   L0
    LINENUMBER 129 L0
    ALOAD 0
    INVOKEVIRTUAL com/shopify/reactnative/flash_list/AutoLayoutView.getParent ()Landroid/view/ViewParent;
    ASTORE 1
   L1
    LINENUMBER 130 L1
   L2
    ALOAD 1
    IFNULL L3
   L4
    LINENUMBER 131 L4
    ALOAD 1
    INSTANCEOF android/widget/ScrollView
    IFNE L5
    ALOAD 1
    INSTANCEOF android/widget/HorizontalScrollView
    IFEQ L6
   L5
    LINENUMBER 132 L5
    ALOAD 1
   L7
    CHECKCAST android/view/View
    ARETURN
   L6
    LINENUMBER 134 L6
    ALOAD 1
    INVOKEINTERFACE android/view/ViewParent.getParent ()Landroid/view/ViewParent; (itf)
    ASTORE 1
   L8
    LINENUMBER 130 L8
    GOTO L2
   L3
    LINENUMBER 136 L3
    ACONST_NULL
    ARETURN
   L9
```

This is bytecode with the sequence:

```
   L0
    LINENUMBER 129 L0
   L1
    LINENUMBER 130 L1
   L2
    LINENUMBER 129 L2
    ALOAD 0
    INVOKEVIRTUAL com/shopify/reactnative/flash_list/AutoLayoutView.getParent ()Landroid/view/ViewParent;
    GETSTATIC com/shopify/reactnative/flash_list/AutoLayoutView$getParentScrollView$1.INSTANCE : Lcom/shopify/reactnative/flash_list/AutoLayoutView$getParentScrollView$1;
    CHECKCAST kotlin/jvm/functions/Function1
    INVOKESTATIC kotlin/sequences/SequencesKt.generateSequence (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Lkotlin/sequences/Sequence;
    ASTORE 1
   L3
    LINENUMBER 130 L3
    NOP
   L4
    ICONST_0
    ISTORE 2
   L5
```

Additionally, you can decompile the bytecode to Java.

For `while`, Java code looks the following:

```
for(ViewParent autoLayoutParent = this.getParent(); autoLayoutParent != null; autoLayoutParent = autoLayoutParent.getParent()) {  
   if (autoLayoutParent instanceof ScrollView || autoLayoutParent instanceof HorizontalScrollView) {  
      return (View)autoLayoutParent;  
   }  
}  
  
return null;
```

And for the sequence, it looks like this:
```
Sequence $this$firstOrNull$iv = SequencesKt.generateSequence(this.getParent(), (Function1)null.INSTANCE);  
int $i$f$firstOrNull = false;  
Iterator var3 = $this$firstOrNull$iv.iterator();  
  
Object var10000;  
while(true) {  
   if (!var3.hasNext()) {  
      var10000 = null;  
      break;  
   }  
  
   Object element$iv = var3.next();  
   ViewParent it = (ViewParent)element$iv;  
   int var6 = false;  
   if (it instanceof ScrollView || it instanceof HorizontalScrollView) {  
      var10000 = element$iv;  
      break;  
   }  
}  
  
if (!(var10000 instanceof View)) {  
   var10000 = null;  
}  
  
return (View)var10000;
```

I haven't dug deeper to see the Java bytecode but it's fair to say the sequence will be slower. However, since it compiles to `while`, the results will be most likely similar and it's more readable. I'd prefer `generateSequence` in case the readability win outweighs the small performance hit but in this case, even the `while` alternative is pretty easy to read, so we can keep it as-is 👍

android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutView.kt
```diff
@@ -66,6 +70,73 @@ class AutoLayoutView(context: Context) : ReactViewGroup(context) {
...
+        val parentScrollView = getParentScrollView()
+        if (disableAutoLayout || parentScrollView == null) {
+            return
+        }
+        val isAutoLayoutEndVisible = if (alShadow.horizontal) right <= parentScrollView.width else bottom <= parentScrollView.height
+        if (!isAutoLayoutEndVisible) {
+            return
+        }
+        val autoLayoutParent = parent as? View
+        val footer = getFooter();
```
fortmarek, line 84: nit:
```suggestion
        val footer = getFooter()
```

ios/Sources/AutoLayoutView.swift
```diff
@@ -76,6 +79,10 @@ import UIKit
         )
     }
 
+    func getScrollView() -> UIScrollView? {
```
fortmarek, line 82:
```suggestion
    func scrollView() -> UIScrollView? {
```

ios/Sources/AutoLayoutView.swift
```diff
@@ -169,15 +177,20 @@ import UIKit
...
             }
+            updateLastMaxBoundOverall(currentCell: cellContainer, nextCell: nextCell)
         }
 
         lastMaxBound = maxBoundNextCell
         lastMinBound = minBound
     }
 
+    private func updateLastMaxBoundOverall(currentCell: CellContainer, nextCell: CellContainer) {
+        lastMaxBoundOverall = max(lastMaxBoundOverall, horizontal ? currentCell.frame.maxX : currentCell.frame.maxY, horizontal ? nextCell.frame.maxX : nextCell.frame.maxY)
```
fortmarek, line 191: nit: this is a bit hard to read, let's split it in multiple lines:
```suggestion
        lastMaxBoundOverall = max(
            lastMaxBoundOverall, 
            horizontal ? currentCell.frame.maxX : currentCell.frame.maxY, 
            horizontal ? nextCell.frame.maxX : nextCell.frame.maxY
       )
```

ios/Sources/AutoLayoutView.swift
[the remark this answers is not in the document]
fortmarek: Agreed, we should add them soon as it makes changes to this file painful and one does not have much confidence they are not breaking anything here.

ios/Sources/AutoLayoutView.swift
```diff
@@ -215,4 +228,44 @@ import UIKit
...
+            superview?.frame.size.width += diff
+        } else {
+            footer.frame.origin.y += diff
+            frame.size.height += diff
+            superview?.frame.size.height += diff
+        }
+    }
+
+    private func footerDiff() -> CGFloat {
+        if subviews.count == 0 {
```
fortmarek, line 258:
```suggestion
        if subviews.isEmpty {
```

`isEmpty` is generally much faster than `count` and more readable 🙂

fortmarek (approved):
Some small nits but already good to merge from my side ✅

Issue #527, opened 2022-07-14
recycled views also recycle prior useState values


sterlingwes:
## Current behavior

Cell components with a `useState` are rendering the state of prior cells, presumably due to the memory cache or previously mounted component reuse (?)

See this repro in the Twitter Timeline fixture app where a "like" of the first tweet shows up on a later tweet:

https://user-images.githubusercontent.com/1047502/178994225-09157b84-ae22-41c5-8e62-b2adae0aa17b.mp4

## Expected behavior

Prior useState values should not affect recycled cells

## To Reproduce

See code: https://github.com/sterlingwes/flash-list/compare/main...sterlingwes:flash-list:usestate-bug-repro

## Platform:

- [x] iOS
- [x] Android

## Environment

1.1.0

sterlingwes:
There's two workarounds for this I figure:
1. don't use `useState`s in anything returned by `renderItem` (probably a good performance idea but kind of unreasonable if the sales pitch is "drop-in replacement for FlatList")
2. add a useEffect to reset any state:

```jsx
const MyRecycledComponent = (props) => {
  const lastKey = useRef(props.myItemKey)
  const [myState, setMyState] = useState(props.myInitialState)
  useEffect(() => {
    if (props.myItemKey !== lastKey.current) {
      lastKey.current = props.myItemKey
      setMyState(props.myInitialState)
    }
  }, [props])
  return <View><Text>{myState}</Text></View>
}
```

If these sound like reasonable suggestions I can open a PR to add to Known Issues in the docs LMK

huduarte:
> 

sorry ;D in my case, extraData works fine

huduarte:
I'd like to re-render only the item I've selected, but since that doesn't work for me, I'm re-rendering the entire list. Maybe it really is a different usecase or else I'm screwing up

naqvitalha:
@sterlingwes The useEffect solution will lead to two updates per render. Isn't ideal as it might cut down JS FPS. I think you can just setState during render like mentioned here: https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

I don't really think it's an issue. It can also happen if you update FlatList data without having stable keys. A component should, ideally, always handle updates.

Instead of known issues, we should add this to the common pitfalls list under usage.

fortmarek:
> I expected the native view representation to be recycled not the react component itself, but maybe I didn't read the docs well enough.

[...]

PR #529, opened 2022-07-14
docs: add known issue for useState

sterlingwes:
## Description

Interim measure related to #527 

The useState issue seems like a pretty big gotcha, so until the underlying issue is resolved it seems worth calling out as a known issue.

- [ ]

## Checklist

### DIFF
--- a/documentation/docs/known-issues.md
+++ b/documentation/docs/known-issues.md
@@ -15,25 +15,29 @@ FlashList and FlatList have very different internal. While the API is almost the
 
 Please note most lists do mount with deterministic sizes so make sure to check if you really need workarounds.
 
-### 2) `onEndReached` event doesn't have `distanceFromEnd`
+### 2) `useState` values will be recycled
+
+If your component hierarchy returned from `renderItem` makes use of `useState`, you will find that state values set by components in other cells of the list may be reflected in cells that use the recycled component. There are two workarounds for this, see [the Github issue](https://github.com/Shopify/flash-list/issues/527#issuecomment-1184463150).
+
+### 3) `onEndReached` event doesn't have `distanceFromEnd`
 
 This value is reported as 0. We don't have plans to implement this right now. Please provide feedback if this is important to you.
 
-### 3)`scrollToIndex` doesn't have `viewOffset, viewPosition` support
+### 4)`scrollToIndex` doesn't have `viewOffset, viewPosition` support
 
 The list only guarantees to bring the given index in view port (as close to the top as possible). There is no support for additional customization.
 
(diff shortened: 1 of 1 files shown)


documentation/docs/known-issues.md
```diff
@@ -15,25 +15,29 @@ FlashList and FlatList have very different internal. While the API is almost the
 
 Please note most lists do mount with deterministic sizes so make sure to check if you really need workarounds.
 
-### 2) `onEndReached` event doesn't have `distanceFromEnd`
+### 2) `useState` values will be recycled
+
+If your component hierarchy returned from `renderItem` makes use of `useState`, you will find that state values set by components in other cells of the list may be reflected in cells that use the recycled component. There are two workarounds for this, see [the Github issue](https://github.com/Shopify/flash-list/issues/527#issuecomment-1184463150).
```
fortmarek, line 20: I don't expect us to solve it as it's expected. I think it's worth creating a dedicated page in `Fundamentals` called e.g. `Using state` where we would detail potential workarounds and reasoning behind it. In this case, it might be better if we write it up most likely - wdyt @naqvitalha?

sterlingwes: that sounds great!

took a stab at roughing in a smaller blurb for you in the interim

documentation/docs/fundamentals/performant-components.md
```diff
@@ -7,7 +7,26 @@ While `FlashList` does its best to achieve high performance, it will still perfo
...
 ## Recycling
 
-One important thing to understand is how `FlashList` works under the hood. When an item gets out of the viewport, instead of being destroyed, the component is re-rendered with a different `item` prop. When optimizing your item component, try to ensure as few things as possible have to be re-rendered and recomputed when recycling.
+One important thing to understand is how `FlashList` works under the hood. When an item gets out of the viewport, instead of being destroyed, the component is re-rendered with a different `item` prop. For example, if you make use of `useState` in a reused component, you may see state values that were set for that component when it was associated with a different item in the list, and would then need to reset any previously set state when a new item is rendered:
+
+```tsx
+const MyItem = ({ item }) => {
+  const lastItemId = useRef(item.someId);
+  const [liked, setLiked] = useState(item.liked);
+  if (item.someId !== lastItemId.current) {
```
naqvitalha, line 16: Let's also link the react docs mentioning this as being the alternative `getDerivedStateFromProps`. I also think we should mention that useState should be used ONLY if necessary and not having states is preferable.

naqvitalha (dismissed):

naqvitalha (approved):

PR #534, opened 2022-07-18
fix: podspec double quotes around version string

jlepp:
## Description

Fixes bug in podspec where version string won't be interpolated correctly.


fortmarek (approved):
Thanks for the update 🙌

PR #548, opened 2022-08-04
made first getDerivedStateFromProps param readonly

Chermaley:
## Description

Types of property 'getDerivedStateFromProps' are incompatible.

image

The problem is in getDerivedStateFromProps first param type, react expects arg of type Readonly.

image
image

My solution is to wrap first arg with Readonly.


fortmarek (approved):
Thanks for the PR, makes sense 🙌

PR #549, opened 2022-08-04
Fabric support on iOS

j-piasecki:
## Description

This PR adds Fabric integration to the FlashList on iOS. Here are some more important highlights:
- I've merged `CellContainer` with `CellContainerComponentView` as keeping a `CellContainer` in every `CellContainerComponentView` didn't look like a very good idea performance-wise
- In case of `AutoLayoutView`, I stayed with wrapping it with the Fabric Component (`AutoLayoutViewComponentView`). I didn't want to accidentally break the functionality converting in to `obj-c`. It may be possible that they can be merged via inheritance but I'm not that knowledgeable in interop between `swift` and `obj-c` and couldn't get it to work
- Event dispatchers on Fabric are auto generated so I added a field to `AutoLayoutView` containing a closure, which in turn is set by the `AutoLayoutViewComponentView`. This way the code in swift has access to the generated `obj-c++` method.
- Due to the view flattening and the fact that the `AutoLayoutView` needs to be wrapped with respective `ComponentView`, the view hierarchy is different on Fabric. I added a `getSubviews()` method that has a different implementation when building for the new architecture.
- I've added a separate `fabricfixture` app to test Fabric integration without changing the original one.

I have tested the `TwitterBenchmark` and the performance there didn't seem to differ between architectures.
Fabric:
| Device | Avg FPS | Min FPS | Max FPS | Max blank area | Cumulative blank area |
|-|-|-|-|-|-|
| iPhone 13 mini | 60.1 | 59.9 | 60.9 | 0 | 0 |
| iPhone 6s | 59.1 | 54.9 | 60.3 | 0 | 0 |

Paper:
| Device | Avg FPS | Min FPS | Max FPS | Max blank area | Cumulative blank area |
|-|-|-|-|-|-|
| iPhone 13 mini | 60.1 | 59.9 | 60.6 | 0 | 0 |
| iPhone 6s | 58.1 | 54.6 | 60 | 0 | 0 |

The `fabricfixture` app is pretty basic at the moment (only `TwitterBenchmark` and commented out `Paginated list`) due to the compatibility of other libraries.

- [ ]

## Checklist


j-piasecki:
Closed in favor of https://github.com/Shopify/flash-list/pull/550.

PR #564, opened 2022-08-15
deps: add recyclerlistview to peerDependencies

alantoa:
## Description
Hi maintainers! 👋  

- sometimes we already have a lib in our project and hope it doesn't need to be reinstalled.
- sometimes I need to patch for [recyclerlistview](https://github.com/Flipkart/recyclerlistview) until it merged, If it's inside this package, it will be bad for processing.

- check the `package.json`'s `dependencies`.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
   "peerDependencies": {
     "@babel/runtime": "*",
     "react": "*",
-    "react-native": "*"
+    "react-native": "*",
+    "recyclerlistview": ">=4.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.18.5",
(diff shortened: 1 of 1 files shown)


fortmarek:
We used to have `recycler-list-view` as a peer dependency but the assumption was that simplifying the usage and installation of `flash-list` is worth the trade-off here as we expect people to usually have only one or the other. 

I'm personally for keeping it as-is and closing this PR. @naqvitalha what are your thoughts?

alantoa:
Hi @fortmarek, thanks for your reply!
I think adding `recyclerlistview ` to `peerDependencies` doesn't affect the `FlashList`'s simplifying the usage and installation. 

Some conditions of peer dependencies: 

- If the user does not have installed `recyclerlistview`: it will be installed according to `dependencies`'s `recyclerlistview`.
- If the user has installed `recyclerlistview`: Its version must be >= 4, otherwise, it will still install inside flashlist.

PR #567, opened 2022-08-16
Fix Prop types on AutoLayoutView to support children not being an intrinsic prop

hbiede:
## Description

When attempting to type check `FlashList` in an Expo 46 environment, TypeScript throws a pair of errors in `AutoLayoutView` due to `children` not being a defined prop of itself nor `AutoLayoutViewNativeComponent`.

Note: If you, like me, thought `"exclude": ["node_modules"]` would fix this, I recommend taking a look at the [tsconfig reference page](https://www.typescriptlang.org/tsconfig#exclude) so we can commiserate on what `exclude` actually means...

Current error:
Screen Shot 2022-08-16 at 3 53 35 PM

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 - Fixes type checking error in `AutoLayoutView` due to `children` not being an explicit type
+  - https://github.com/Shopify/flash-list/pull/567
 
 ## [1.2.1] - 2022-08-03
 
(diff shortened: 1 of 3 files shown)


fortmarek (approved):
Not sure about `excludes` here but I'm fine with adding `children` prop explicitly here 👍

hbiede:
Is there an expected timeline for release? This is a build blocker for us on the story to take up using flash list

hbiede:
@fortmarek Following up on this

fortmarek:
Sorry @hbiede for the late response – we'll do another release today to unblock you 👍

PR #576, opened 2022-08-31
Solves the header remount when switching from an empty data array #575

### LINKED ISSUE #575: Header component is unmounted and remounted when data changes
## Current behavior

The header component is unmounted and remounted after data changes from an empty array to an array with records.

## Expected behavior

The header should render with empty and not empty data, but without unmounting the component 

## To Reproduce

```js
import { StatusBar } from "expo-status-bar";
import { StyleSheet, Text, View } from "react-native";
import { FlashList } from "@shopify/flash-list";
import React from "react";

const DATA = [];

const DATA1 = [
  {
    title: "First Item",
  },
  {
    title: "Second Item",
  },
];

const Header = () => {
  React.useEffect(() => {
    console.log("header mount");
    return () => console.log("header unmount");
  }, []);
  return (
    <View style={{ paddingTop: 50 }}>
      <Text>Header</Text>
    </View>
  );
};

const keyExtractor = (item) => item.title;

  const [data, setData] = React.useState(DATA);

  React.useEffect(() => {
    setData(DATA1);
  });

  return (
    <View style={styles.container}>
      <FlashList
        ListHeaderComponent={<Header />}
        ListEmptyComponent={<Text>EMPTY</Text>}
        keyExtractor={keyExtractor}
        data={data}
        renderItem={({ item }) => <Text>{item.title}</Text>}
        estimatedItemSize={200}
      />
      <StatusBar style="auto" />
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: "#fff",
    alignItems: "center",
    justifyContent: "center",
  },
});

```

## Platform:

- [ ] iOS
- [ ] Android

## Environment

 @shopify/flash-list v1.2.1
 
Reproduced with expo  "~46.0.9" and a bare react native project

msantang78:
## Description

Fixes #575

This PR solves the header unmount when switching from an empty data array.

But, being unfamiliar with the code of the package, I'm not sure if it can potentially introduce any regression.

- [ ]

## Checklist

### DIFF
--- a/src/FlashList.tsx
+++ b/src/FlashList.tsx
@@ -427,7 +427,7 @@ class FlashList<T> extends React.PureComponent<
     return (
       <>
         <PureComponentWrapper
-          enabled={children.length > 0 || this.isEmptyList}
+          enabled={true}
           contentStyle={this.props.contentContainerStyle}
           horizontal={this.props.horizontal}
           header={this.props.ListHeaderComponent}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 - Fixes type checking error in `AutoLayoutView` due to `children` not being an explicit type
   - https://github.com/Shopify/flash-list/pull/567
+- Fixes the header re-mount when switching from an empty data array
+  - https://github.com/Shopify/flash-list/pull/576
 
 ## [1.2.1] - 2022-08-03
 


src/FlashList.tsx
```diff
@@ -427,7 +427,7 @@ class FlashList<T> extends React.PureComponent<
     return (
       <>
         <PureComponentWrapper
-          enabled={children.length > 0 || this.isEmptyList}
+          enabled={true}
```
naqvitalha, line 430: Instead of this change, we can add this.isListLoaded as the first OR condition in both footer and header. Again, this is something already fixed in one of the branches.

msantang78: perfect, thanks!

PR #625, opened 2022-10-05
Expose columnIndex to MasonryFlashList.renderItem via extraData.

jamesism:
## Description

Adds an implicit `columnIndex: number` to `extraData` when using `MasonryFlashList` in order to expose which column information to the presentation layer.

### DIFF
--- a/src/MasonryFlashList.tsx
+++ b/src/MasonryFlashList.tsx
@@ -179,6 +179,10 @@ const MasonryFlashListComponent = React.forwardRef(
                     ...innerArgs,
                     item: innerArgs.item.originalItem,
                     index: innerArgs.item.originalIndex,
+                    extraData: {
+                      ...innerArgs.extraData,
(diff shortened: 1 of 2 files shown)


src/MasonryFlashList.tsx
```diff
@@ -179,6 +179,10 @@ const MasonryFlashListComponent = React.forwardRef(
                     ...innerArgs,
                     item: innerArgs.item.originalItem,
                     index: innerArgs.item.originalIndex,
+                    extraData: {
```
naqvitalha, line 182: Why do you need the column number? Columns are abstracted because if we build a way in the future to change column span then `columnIndex` will need to be deprecated. If an item spans across columns then API should be `startColumn` and `endColumn`.
Also, changing extraData ref can cause unwanted rendering as we guarantee forwarding the same reference to components but this is going to change that.

PR #626, opened 2022-10-05
Content padding support

### LINKED ISSUE #621: Support all padding props in contentContainerStyle
### Description
FlashList doesn't support horizontal padding on vertical lists which makes it difficult to apply uniform padding when numColumns value is more than 1. It's even harder in masonry because items can be rearranged.

## Current behavior
A warning is raised if padding is applied on vertical list.

## Expected behavior

Padding should work.

## Platform:

- [x] iOS
- [x] Android

naqvitalha:
## Description

resolves #621 

We've had to ignore horizontal padding on vertical lists and the opposite for horizontal ones due to layout complexities. This PR solves the problem and avoids any performance regression. 

### Motivation
Doing this now was important because, in regular FlashList, padding can simply be added to items but in masonry, we change the order of items and it's quite complex to add something as simple as content padding. Developers need to render and update items based on their columns.

- [ ] Try playing with `contentContainerStyle` in any of the sample. 
- [ ] Check performance regression (there isn't any but doesn't hurt to check)

## Checklist


naqvitalha:
@fortmarek Can you take a look at this when you have time? It'll be very helpful for the community

src/utils/ContentContainerUtils.ts
```diff
@@ -0,0 +1,92 @@
+import { ViewStyle } from "react-native";
+import { Dimension } from "recyclerlistview";
+
+import { ContentStyle } from "../FlashListProps";
+
+export interface ContentStyleExplicit {
```
fortmarek, line 6: What do you think of naming this `InternalContentStyle`? I understand where `explicit` is coming from but sounds odd. Ok to keep as-is though.

src/MasonryFlashList.tsx
```diff
@@ -430,8 +438,13 @@ const updateViewTokens = (tokens: ViewToken[]) => {
...
     const token = tokens[i];
     if (token.index !== null && token.index !== undefined) {
-      token.index = token.item.originalIndex;
-      token.item = token.item.originalItem;
+      if (token.item) {
+        token.index = token.item.originalIndex;
+        token.item = token.item.originalItem;
+      } else {
+        token.index = null;
+        token.item = undefined;
```
fortmarek, line 446: Why `index` is set to `null` and `item` to `undefined`?

naqvitalha: Unfortunately, that's the type specified in ViewToken. Changing it can be a breaking change for many.

src/utils/ContentContainerUtils.ts
```diff
@@ -0,0 +1,92 @@
...
+    padding,
+    paddingVertical,
+    paddingHorizontal,
+    backgroundColor,
+  } = (contentContainerStyleSource ?? {}) as ViewStyle;
+  contentStyle.paddingLeft = Number(
+    paddingLeft || paddingHorizontal || padding || 0
+  );
+  contentStyle.paddingRight = Number(
+    paddingRight || paddingHorizontal || padding || 0
```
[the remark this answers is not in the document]
naqvitalha: This sequence was already there. I'm only casing to number now. I don't think it's worth the change and can also be a breaking change.

src/GridLayoutProviderWithProps.ts
```diff
@@ -57,10 +59,32 @@ export default class GridLayoutProviderWithProps<T> extends GridLayoutProvider {
...
+    this._hasExpired =
+      this._hasExpired ||
+      newInsetValues.height !== this.renderWindowInsets.height ||
+      newInsetValues.width !== this.renderWindowInsets.width;
+    this.renderWindowInsets = newInsetValues;
     this.props = props;
+    return this;
+  }
+
+  public get hasExpired() {
```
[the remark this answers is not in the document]
naqvitalha: Done

PR #654, opened 2022-10-31
Fix style validation

### LINKED ISSUE #651: Flashlist produces warning when passed to StyledComponent
## Current behavior
Our project is using [StyledComponents](https://styled-components.com/docs/basics). Now, while migrating to Flashlist I found that it is producing warning, even though I didn't pass any styles to it.
```
import styled from 'styled-components/native'

const List = styled(FlashList).attrs(({ theme }) => ({
  ListHeaderComponentStyle: {
    backgroundColor: theme.color.background,
    borderBottomWidth: 1,
    borderColor: theme.color.separator,
  },
}))({})
```

image

## Expected behavior

No warning if styles are empty

## To Reproduce

see code snipped above

## Platform:

- [x] iOS
- [x] Android

## Environment
    "@shopify/flash-list": "1.1.0"

kubilaysalih:
## Description

Fixes #651 

Style prop can be an array. So we need to validate every object in array.
Like `createAnimatedComponent` always passes a blank style object, `styled-components`passes an array with an empty object.

- [ ]

## Checklist


kubilaysalih:
I've signed CLA.

fortmarek:
I am a bit hesitant to merge this as it fixes a case for a third-party library. Is it possible to change instead what `styled` passes to `FlashList`?

kubilaysalih:
Actually this solution has nothing to do with a third party library. `styled-components` is an example that produces this warning. Any other higher order component that passes style property as an array can cause this. Because the problem is passing style property as an array.

For example the following usage will not cause a problem and we will not see a warning.
`
  <FlashList {...props} style={{}}  />
`

But if you want to use it likes this, things will change.
`
  <FlashList {...props} style={[{}]}  />
`

First example does not give a warning, why does the second? I think they are kinda similar components that don't have a style. Of course, if there is a valid style property as an object or as an array of objects, we need to warn developer, because style property is not supported yet.

The summary of the PR is that the style property can be an array, and we need to consider this as well.

src/FlashList.tsx
```diff
@@ -126,10 +126,14 @@ class FlashList<T> extends React.PureComponent<
       throw new CustomError(ExceptionList.columnsWhileHorizontalNotSupported);
     }
 
+    const styles = Array.isArray(this.props.style) ? this.props.style : [this.props.style]
+
     // `createAnimatedComponent` always passes a blank style object. To avoid warning while using AnimatedFlashList we've modified the check
-    if (Object.keys(this.props.style || {}).length > 0) {
+    // `style` prop can be an array. So we need to validate every object in array. Check: https://github.com/Shopify/flash-list/issues/651
+    if(styles.some((e) => Object.keys(e || {}).length > 0)) {
```
fortmarek, line 133: We can use [flatten](https://reactnative.dev/docs/stylesheet#flatten) instead. [This](https://github.com/Shopify/flash-list/pull/654/files#diff-f8655d3e55acbb01f8cca2007bd8e740be197492eecb49d145dbdf476a423b0fR129) line can then be reverted. This will also work for `style={[[{}]]}` or whatever combination of empty arrays you would use.

Since this check is relevant only in `DEV`, we should also add a check for that first.


```suggestion
    if (__DEV__ && Object.keys(StyleSheet.flatten(this.props.style ?? {})).length > 0) {
```

fortmarek (commented):
> Actually this solution has nothing to do with a third party library. styled-components is an example that produces this warning. Any other higher order component that passes style property as an array can cause this. Because the problem is passing style property as an array.

You are correct, since passing an array of styles is valid, we should check for this.

Sorry for the longer response time, I have some minor feedback after which we can get this merged 🙂

fortmarek (approved):
:shipit:

PR #660, opened 2022-11-03
Upgrade RLV to v4.2.0

naqvitalha:
## Description

Fixes couple of warnings and some precision issues with column width computation. Here's the full changelog: https://github.com/Flipkart/recyclerlistview/compare/4.1.2...4.2.0

I think the precision issues that were fixed in RLV can help with contentPaddingSupport so, pulling in RLV is a good idea. Please note that v4.2.0 was released a month ago and has no reported regressions. 

General regression, I don't expect anything to break.

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6645,10 +6645,10 @@ recast@^0.20.4:
     source-map "~0.6.1"
     tslib "^2.0.1"
 
-recyclerlistview@4.1.2:
-  version "4.1.2"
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Add content padding support to FlashList
```
naqvitalha, line 10: This update to the changelog was missed. Fixing it in this PR

fortmarek (approved):

PR #692, opened 2022-11-29
Update performant-components.md

nandorojo:
Small docs change.


fortmarek (commented):
Is there are any reason for this change? 🤔

nandorojo:
Yes. The current docs are incorrect. They use `memo()` inside of render code. As a result, it creates a brand new component instance on every render. This defeats the purpose.

These docs move the use of `memo` to outside of the component.

I made a [sandbox](https://codesandbox.io/s/hidden-shadow-4boid4?file=/src/App.js) to show why the current docs break the React rules of memoization.

To test it, open the `console` in the bottom right, and then click the `Rerender` button. Notice that the component wrapped with `memo` still logs.

fortmarek (approved):
Thanks for the explanation!

We can merge this once you sign the CLA, let me know if you have any issues with that 🙂

naqvitalha:
Closing this as it has been pending for a long time.

Issue #751, opened 2023-02-02
Inverted scroll super laggy on Android


devoren:
I want to use a  flashlist to show a inverted list. But if I use inverted prop on android it works very slow.
Default Flashlist: drops 5-10 FPS in dev mode (UI 60FPS)
Inverted Flashlist: drops 30-40 FPS in dev mode (UI also 20-40FPS)

```
const renderItem: ListRenderItem<IMessage> = ({ item, index }) => {
	return (
		<MessageItem id={item.id} message={item.message} date={item.date} />
	)};

 const keyExtractor = ({ id, date }: IMessage) => `${id}_${date}`;

 return (
	<SafeAreaView style={styles.container}>
		<FlashList
			ref={list}
			data={messages}
			keyExtractor={keyExtractor}
			renderItem={renderItem}
			contentContainerStyle={{
				paddingVertical: SCALE.XS,
			}}
			showsVerticalScrollIndicator={false}
                        estimatedItemSize={100}
		/>
	</SafeAreaView>
  );
```

## Current behavior

[inverted.webm](https://user-images.githubusercontent.com/99968085/216315189-47eef7bc-5d3b-4390-a06c-54ce1ad48b5a.webm)

## Expected behavior
UI 60 FPS

## To Reproduce

The flatlist also has the same behavior but after using [this](https://github.com/facebook/react-native/issues/30034#issuecomment-1411922104) patch UI stable 60fps but not working with FlashList
## Platform:

- [ ] iOS
- [+] Android

## Environment

1.4.1

CaptainJeff:
Same issue

naqvitalha:
I'm looking into this.

devoren:
@CaptainJeff What if we change this  code 

<img width="477" alt="image" src="https://user-images.githubusercontent.com/99968085/216997028-529714e1-3743-486f-bd1b-0a1eca36add4.png">
 
 to this code ( flatlist solution ):
 Platform.OS === 'android' ? { scaleY: -1 } : { transform: [{scaleY: -1}] }

CaptainJeff:
Yeah @devoren that's pretty much my current temporary solution. I was doing a patch-package on the react-native library for `transformStyle` but just switched it to patch just this dependency. It does seem to work.

@naqvitalha maybe it makes sense to update this in the project? I just opened a PR https://github.com/Shopify/flash-list/pull/755

CaptainJeff:
[...]

PR #755, opened 2023-02-06
fix: replacing transform scale style for android

CaptainJeff:
Updating the transform style for android from `{ transform: [{ scaleX: -1 }] }` to `{ scaleY: -1 }`

Fixes https://github.com/Shopify/flash-list/issues/751

This issue seems to mostly impact android 13 phones where inverted is `true`. Can see a video of it happening on https://github.com/Shopify/flash-list/issues/751

On an android 13 with a decently sized list you should be able to scroll and fetch extraData without frame drops or ANR issues

## Checklist

### DIFF
--- a/src/FlashList.tsx
+++ b/src/FlashList.tsx
@@ -5,6 +5,7 @@ import {
   LayoutChangeEvent,
   NativeSyntheticEvent,
   StyleSheet,
+  Platform,
   NativeScrollEvent,
 } from "react-native";
 import {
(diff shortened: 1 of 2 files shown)


src/FlashList.tsx
```diff
@@ -69,7 +70,10 @@ class FlashList<T> extends React.PureComponent<
   private rlvRef?: RecyclerListView<RecyclerListViewProps, any>;
   private stickyContentContainerRef?: PureComponentWrapper;
   private listFixedDimensionSize = 0;
-  private transformStyle = { transform: [{ scaleY: -1 }] };
+  private transformStyle =
```
naqvitalha, line 73: We keep our platform-specific code inside the native/ directory. Can you please move it there?
This also doesn't look like an official spec. Does FlatList also do this internally or is this optimization only part of a certain patch?

naqvitalha:
Yes, you are right. We're going to close this now.

harisgaul:
tes

PR #788, opened 2023-03-15
Apply layout corrections only to consecutive cells

### LINKED ISSUE #600: [bug][ios]: using getItemType causes blanks and overlapping items
## Current behavior
Using `getItemType` causes blanks and overlapping items.

I reproduced a bug where using getItemType causes blanks and overlapping items. This does only happen when using different types but the elements having the same height. In my Expo Snack, just using two types (e.g "header" and "row" does not break), since header and row have different heights. But I used different types (in my real app) because the items have different icons and styling, so I tried to add more types in order to optimize renders. While I could just use two types to "fix" this issue, getItemType should never cause this blanks, even when the items have the same height.

https://user-images.githubusercontent.com/504909/190257015-3e38ea4e-09d1-47b4-a6b1-4b246881a0e2.mp4

Bildschirmfoto_2022-09-14_um_21 54 43

I was only able to reproduce this on iOS, but honestly haven't tried android enough. It is not breaking on web.

## Expected behavior

Should not cause blanks and overlapping items

## To Reproduce

https://snack.expo.dev/@hirbod/getitemtype-failing

## Platform:

- [X] iOS
- [ ] Android

## Environment

Happens on iOS (bare react-native and expo, 1.1.0 and 1.2.1)

naqvitalha:
## Description

resolves #600 

This issue was reported long ago but we couldn't identify the root cause. We recently figured out that the issue happens when you have two items of different types next to each other but not in consecutive order from data perspective.

For example, assume View1, View2, View3, View4, View7, View8 are present in the view hierarchy. From the parent view's perspective View4 and View7 are next to each other but they're actually not because we should have View5 and View6 in between. The reason for two views are missing could be many one of them being slow rendering due to expensive components and View7 / View8 are already in view tree because they're of different type so they might not be getting recycled at the same time. 
In the mentioned case AutoLayout will also think View4 and 7 are neighbours with a gap and it'll pull 7 up pushing it out of sync with react layout. This may or may not get fixed based on future onLayout operations. This is why we can see overlaps or gaps.

This PR should resolve the issue by making sure AutoLayout understands whether or not views are consecutive.

Try the list sample on a slow device. Scroll down so that you see item 98 and 99 once. Scroll to the top and quickly scroll down. You will notice red items popping in and out of places where they should never be.

## Checklist

### DIFF
--- a/android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutShadow.kt
+++ b/android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutShadow.kt
@@ -25,23 +25,27 @@ class AutoLayoutShadow {
         for (i in 0 until sortedItems.size - 1) {
             val cell = sortedItems[i]
             val neighbour = sortedItems[i + 1]
+            // Only apply correction if the next cell is consecutive.
+            val isNeighbourConsecutive = cell.index == neighbour.index + 1
             if (isWithinBounds(cell)) {
                 if (!horizontal) {
                     maxBound = kotlin.math.max(maxBound, cell.bottom);
                     minBound = kotlin.math.min(minBound, cell.top);
                     maxBoundNeighbour = maxBound
-                    if (cell.left < neighbour.left) {
-                        if (cell.right != neighbour.left) {
-                            neighbour.right = cell.right + neighbour.width
-                            neighbour.left = cell.right
+                    if (isNeighbourConsecutive) {
+                        if (cell.left < neighbour.left) {
+                            if (cell.right != neighbour.left) {
+                                neighbour.right = cell.right + neighbour.width
+                                neighbour.left = cell.right
+                            }
+                            if (cell.top != neighbour.top) {
+                                neighbour.bottom = cell.top + neighbour.height
+                                neighbour.top = cell.top
+                            }
+                        } else {
+                            neighbour.bottom = maxBound + neighbour.height
+                            neighbour.top = maxBound
                         }
-                        if (cell.top != neighbour.top) {
-                            neighbour.bottom = cell.top + neighbour.height
-                            neighbour.top = cell.top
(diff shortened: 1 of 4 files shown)


android/src/main/kotlin/com/shopify/reactnative/flash_list/AutoLayoutShadow.kt
```diff
@@ -25,23 +25,27 @@ class AutoLayoutShadow {
         for (i in 0 until sortedItems.size - 1) {
             val cell = sortedItems[i]
             val neighbour = sortedItems[i + 1]
+            // Only apply correction if the next cell is consecutive.
+            val isNeighbourConsecutive = cell.index == neighbour.index + 1
```
fortmarek, line 29: Should we apply this logic on iOS, too, or does the problem not manifest there?

naqvitalha: iOS changes are also present in this PR. Check below

ios/Sources/AutoLayoutView.swift
```diff
@@ -5,267 +5,273 @@ import UIKit
...
-        }
-        let autoLayoutEnd = horizontal ? frame.width : frame.height
-        return lastMaxBoundOverall - autoLayoutEnd
-    }
-
-    private func footer() -> UIView? {
-        return superview?.subviews.first(where:{($0 as? CellContainer)?.index == -1})
-    }
+	@objc(onBlankAreaEvent)
+	var onBlankAreaEvent: RCTDirectEventBlock?
```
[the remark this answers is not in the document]
naqvitalha: Yeah apologies for this. I reformatted because indentations were looking off in github. I'll revert ASAP

naqvitalha:
@AlexanderEggers Replied on the issue

PR #819, opened 2023-04-20
Change transform style on Android to fix perf

### LINKED ISSUE #751: Inverted scroll super laggy on Android

naqvitalha:
## Description

resolves #751 

Change transform style to fix inverted performance issue on Android.

Messages sample should perform flawlessly.

## Checklist

### DIFF
--- a/src/native/config/PlatformHelper.android.ts
+++ b/src/native/config/PlatformHelper.android.ts
@@ -2,6 +2,7 @@ import { BaseItemAnimator } from "recyclerlistview";
 
 const PlatformConfig = {
   defaultDrawDistance: 250,
+  // Using rotate instead of scaleY on Android to avoid performance issues. Issue: https://github.com/Shopify/flash-list/issues/751
   invertedTransformStyle: { transform: [{ rotate: "180deg" }] },
 };
 const getCellContainerPlatformStyles = (


package.json
```diff
@@ -28,7 +28,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "up": "bundle install && yarn fixture-up && yarn e2e-up && yarn build && yarn fixture-web-up",
+    "up": "bundle install && yarn fixture-up && yarn e2e-up && yarn build",
```
naqvitalha, line 31: Removed due to some build issue.

src/native/config/PlatformHelper.android.ts
```diff
@@ -2,6 +2,7 @@ import { BaseItemAnimator } from "recyclerlistview";
 
 const PlatformConfig = {
   defaultDrawDistance: 250,
+  invertedTransformStyle: { transform: [{ rotate: "180deg" }] },
```
fortmarek, line 5: Maybe we can add a comment for why we're doing this and link to the original issues. Might be surprising to see this down the road.

fortmarek (approved):

irisjae:
@KYZNCODE-Sebastian-Roese Indeed, checking the upstream patch suggests that we should pass `isInvertedVirtualizedList` to the inner ScrollView when using inverted lists on Android. This can be achieved by adding `overrideProps={{ isInvertedVirtualizedList: true }}` to the FlashList.

saif-o99:
@naqvitalha Hey, Could you please tell which one is better in performance wish, `transform or scale`?
because the current solution creates an issue  for the scroll indicator and i've seen somewhere the solution is `scale` but i don't want to lose any performance.
Thanks.

saif-o99:
@irisjae that works!. Thank you, any idea what does it do internally? is it just styling or may it affect the performance?

irisjae:
Read my previous comment, and the linked commit. tldr, React Native fixed the issue, and this Flash List patch for the issue doesn't uses half of the upstream fix without the other half, which is the prop which I just pointed you to.

> @irisjae that works!. Thank you, any idea what does it do internally? is it just styling or may it affect the performance?

PR #832, opened 2023-04-27
Fix crash if estimatedItemSize is NaN

ritmillio:
## Description

Fixes (issue #)
This PR fixes a crash caused by the estimatedItemSize having a NaN value in some cases. (I came across with the problem when I was developing a Carousel component where I was calculating the estimatedItemSize inside of a useCallback.) The application now checks for NaN values and sets estimatedItemSize to null, which prevents the crash. However, this may cause performance issues, so a warning message is displayed in the console to alert developers about the potential problem.

## Checklist
- Check if the implemented solution prevents crashes caused by NaN estimatedItemSize values.
- Check if the implemented solution prevents crashes caused by NaN estimatedItemSize values.

### Fixed
-estimatedItemSize is NaN check

### DIFF
--- a/src/FlashList.tsx
+++ b/src/FlashList.tsx
@@ -730,6 +730,13 @@ class FlashList<T> extends React.PureComponent<
         );
       }, 1000);
     }
+    if (isNaN(this.props.estimatedItemSize)) {
+      this.props.estimatedItemSize = null;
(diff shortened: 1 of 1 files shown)


naqvitalha:
Closing given the prop is removed in v2.

Issue #633, opened 2022-10-13
Separators not rendered when adding more items to bottom of a list


sahil-ahuja-1:
I'm seeing a bug as follows with ItemSeparatorComponent

I have a list of 10 items with separators in between. I append 10 more items to the list with pagination. The 10th item (previously the last item) doesn't show an item separator between it and the 11th item after adding the additional items to the list. When I scroll far enough down to recycle the view and scroll back up, the separator shows

fortmarek:
Please, provide reproduction steps for this.

mk-nickyang:
@fortmarek I made a reproduction snack here, let me know if there is anything else you need
https://snack.expo.dev/@lucissa/list-separator-not-rendered-when-adding-more-items

anatooly:
Yeap, ItemSeparatorComponent render logic at FlatList not equal as FlashList.
lastEl -1 not have separator from lastEl.
Add separator for position ListHeaderComponent when use inverted prop, strange.

Code-Victor:
Good day, any progress on this?
I'm still facing this issue

aaronabf:
How is this still an open issue for years now?

naqvitalha:
This is resolved in FlashList v2.

PR #845, opened 2023-05-14
fix: item separator issue

Buksirchik:
## Description

Fixes (#633)

_Issue summary:_
The `ItemSeparatorComponent` doesn't render properly in a case when adding items to the bottom (e.g. infinity scroll). It causes because a separator is rendered at the bottom of the cell and when new data is added, previous items don't render due to memoization.

Due to the placement of `ItemSeparatorComponent` has been changed from bottom to top, we have to change the leading and trailing items.
Instead of these statements:
```
const leadingItem = this.props.data[index];
const trailingItem = this.props.data[index + 1];
  ```
  we should use:
  ```
const leadingItem = this.props.data[index - 1];
const trailingItem = this.props.data[index];
  ```
  
<details> 
  <summary>Horizontal inverted FlashList (android, iOS)</summary>

https://github.com/Shopify/flash-list/assets/64099626/0c12cdfa-3040-4eac-8f72-f5a17093c465

https://github.com/Shopify/flash-list/assets/64099626/dcf475a5-f068-49fe-bd00-c2ad505c823f
</details>

<details> 
  <summary>Horizontal FlashList (android, iOS)</summary>

https://github.com/Shopify/flash-list/assets/64099626/1b7a5a75-712b-4ec9-8191-e77ecb14fc9d

https://github.com/Shopify/flash-list/assets/64099626/920983ba-7f34-4678-8b1d-8f45701745dd

</details>

<details> 
  <summary>Vertical FlashList (android, iOS)</summary>

https://github.com/Shopify/flash-list/assets/64099626/85fd191a-20d1-4a9a-ac77-70cbb03f6099

https://github.com/Shopify/flash-list/assets/64099626/cb973507-8aa1-4df9-855d-54e12d78ba2c

</details>

<details> 
  <summary>Vertical inverted FlashList (android, iOS)</summary>

https://github.com/Shopify/flash-list/assets/64099626/c78d1a11-bc36-4357-9741-7edceb4b82ce

https://github.com/Shopify/flash-list/assets/64099626/f476a3e7-2f37-4c7e-a9cd-0a588126dd52

</details>

## Checklist

### CODE REVIEW

**Comment by Buksirchik** [comment]
I have signed the CLA!

**Comment by chj-damon** [comment]
I'm facing the same problem, can this PR be merged soon? @naqvitalha

**Comment by Gondolav** [comment]
Hi, I'm also facing the same problem @fortmarek @naqvitalha

**Comment by irion94** [comment]
I've created a patch from your PR and encountered an issue when the list works with Sticky Headers.

Upon further investigation, I believe a slightly different approach may be beneficial. We can make `trailingItem` undefined, allowing developers more flexibility to decide what should be rendered in this situation.

This change provides greater control and customization options when working with the list. Your feedback on this approach is appreciated.

So, now, when the list is recharged on the bottom - I always add the separator

Thank you for raising this issue. I can confirm its existence, but I believe it's crucial for us to take some time to carefully consider the best approach for addressing it.  ✌️

```
  private separator = (index: number) => {
    // Make sure we have data and don't read out of bounds
    if (this.props.data === null || this.props.data === undefined) {
      return null;
    }

    const leadingItem = this.props.data[index];
    const trailingItem = this.props.data[index + 1];

    const props = {
      leadingItem,
      trailingItem,
      // TODO: Missing sections as we don't have this feature implemented yet. Implement section, leadingSection and trailingSection.
      // https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/Libraries/Lists/VirtualizedSectionList.js#L285-L294
    };
    const Separator = this.props.ItemSeparatorComponent;
    return Separator && <Separator {...props} />;
  };

irion94:
@Buksirchik, I apologize for not providing more details earlier. To illustrate the issue, I've prepared a basic example that showcases the behavior on the release version of the flashlist and with your patch applied.
It's not about the `ItemSeparatorComponent` isn't rendered. Rather, the concern lies in the behavior of the animation between section headers.

In addition, I've included a snippet of my Separator component (although I didn't anticipate it causing the problem).
As we examine the behavior, it appears that the section headers are being calculated incorrectly.

I'm not certain if `/fixture` includes an example of section header usage. Therefore, it might be a good idea to review the behavior of your fix independently  


https://github.com/Shopify/flash-list/assets/34029163/9e413c62-c981-445e-8e1a-89d4ce7c56b6


https://github.com/Shopify/flash-list/assets/34029163/33f3b137-7d9f-4870-a01f-94ac14185daf


```
export const __Separator = ({
	leadingItem,
	trailingItem,
}: TimelineItemSeparatorProps) => {
	if (trailingItem.type === 'type-a' && leadingItem.type === 'type-b')
		return <Separator size={'BIG'} />;

	if (trailingItem.type === 'type-b' && leadingItem.type === 'type-b')
		return <Separator size={'SMALL'} />;

	if (trailingItem.type === 'type-c' && leadingItem.type === 'type-b')
		return <Separator size={'SMALL'} />;

	if (trailingItem.type === 'type-b' && leadingItem.type === 'type-c')
		return <Separator size={'SMALL'} />;
	if (trailingItem.type === 'type-a' && leadingItem.type === 'type-d')
		return <Separator size={'BIG'} />;

	return null;
};

Buksirchik:
@irion94 Could you share your repo with me or how can I reproduce it?
Because I achieved a little bit the same behaviour as yours and it seems everything is ok

<details>
<summary>
main branch
</summary>

https://github.com/Shopify/flash-list/assets/64099626/ecd40314-387c-4c16-8b43-285ca1699c4b

</details>

<details>
<summary>
my patch
</summary>


https://github.com/Shopify/flash-list/assets/64099626/29e2504c-15c8-4bf2-a894-9002af8bd041

</details>

Do you have a custom animation?

Buksirchik:
@irion94, according to your issue, I'll try to invistigate soon. Thanks for passing more details. Also I wann know, do you have any custom animation in your example? Did you only use `stickyHeaderIndices`?

irion94:
> @irion94, according to your issue, I'll try to invistigate soon. Thanks for passing more details. Also I want to know, do you have any custom animation in your example? Did you only use `stickyHeaderIndices`?

I'm using `stickyHeaderIndices` and nothing else. I tried to test your code on the most basic example 😀

> > Upon further investigation, I believe a slightly different approach may be beneficial. We can make `trailingItem` undefined, allowing developers more flexibility to decide what should be rendered in this situation.
> 
> Also I don't understand why do you need it?
> 
> The main point is that `ItemSeparatorComponent` should be shown if we have quantity of items bigger than 1. If we have only 1 item, we shouldn't render `ItemSeparatorComponent`. My opinion on that — we shouldn't pass `undefined` to `trailingItem`. On the other hand you can be right, but it's out of scope of that issue. 😊

Yes, I agree, It would be better to handle it as it is right now, but if we don't find a way out I'll keep my version 😅

Anyway, please confirm if the problem truly occurs and let me know ✌️

Buksirchik:
> Anyway, please confirm if the problem truly occurs and let me know ✌️

@irion94 I confirm issue. I reproduced it.

Buksirchik:
Guys, I've decided to close that PR due to fix isn't good enough. It causes a lot of problem with animations and sticky headers (see problem described by @irion94). It seems the better way to handle separator via [renderItem](https://shopify.github.io/flash-list/docs/usage#renderitem).

PS. I'm sorry about long feedback, it took about 1.5 year :(

PR #890, opened 2023-08-01
Update react-native types to comply to RN 0.72

gvarandas:
## Description

- Update types to complain with the updated `react-native` types shipped as part of `0.72`.
- Add an npm script to verify types are sound without having to build the package.

No actual changes are being shipped as part of it, except for type changes.

- [ ] CI should be ✅ 

**N/A**

## Checklist


gvarandas:
@fortmarek @naqvitalha  this PR might affect which RN versions the library supports, given that users might run into TS issues in case they haven't upgraded. We should carefully consider this before releasing it into the wild.

fortmarek (approved):
LGTM ✅

yarn.lock
```diff
@@ -1739,11 +1747,12 @@
   resolved "https://registry.yarnpkg.com/@types/prop-types/-/prop-types-15.7.4.tgz#fcf7205c25dff795ee79af1e30da2c9790808f11"
   integrity sha512-rZ5drC/jWjrArrS8BR6SIr4cWpW09RNTYt9AMZo3Jwwif+iacXAqgVjm0B0Bv/S1jhDXKHqRVNCbACkJ89RAnQ==
 
-"@types/react-native@0.67.6":
-  version "0.67.6"
-  resolved "https://registry.yarnpkg.com/@types/react-native/-/react-native-0.67.6.tgz#9a7de5feba6065aec9f44f9a1e8f6e55ee5d015c"
-  integrity sha512-NM6atxrefIXMLE/PyQ1bIQjQ/lWLdls3uVxItzKvNUUVZlGqgn/uGN4MarM9quSf90uSqJYPIAeAgTtBTUjhgg==
+"@types/react-native@0.72.2":
```
fortmarek, line 1750: Maybe we can remove this dependency completely now that `react-native` ships with types?

gvarandas: In order to do that, we'd need to declare `react-native` as a devDependency instead. It's a more future-proof alternative, I believe, as I'm not really sure how long the `@types/react-native` lib will be actively maintained now that we have official support for TS in the original repo.

gvarandas:
@fortmarek it really depends on how far we want to go with backwards compatibility. The updated types are more permissive in this case (they add the `'auto'` type for dimension values), so it shouldn't be an issue overall. It is, however, really hard to guarantee that everything will work with super old RN versions.

yarn.lock
fortmarek: > It's a more future-proof alternative, I believe, as I'm not really sure how long the @types/react-native lib will be actively maintained now that we have official support for TS in the original repo.

IIRC, that dependency will no longer be updated from 0.73 onwards. I agree we can keep as-is for this PR, though.

fortmarek:
> The updated types are more permissive in this case (they add the 'auto' type for dimension values), so it shouldn't be an issue overall. It is, however, really hard to guarantee that everything will work with super old RN versions.

If that's the case, I'd ship this. While we don't have an official support window, I don't think we should try to support really old RN versions.

PR #1106, opened 2024-03-20
Fix iOS build and web app build

naqvitalha:
## Description

- This PR removes Flipper which was causing build failures on iOS.
- It also splits the fixture app folders into separate entries:
```
- root
  - fixture
    - react-native
    - expo
```
- Updates the root `package.json` commands to reflect this change and standardizes the commands using the `:` notation to differentiate between targets. For example:
  - `fixture:rn:up` installs the dependencies for the managed `react-native` fixture
  - `fixture:expo:up` installs the dependencies for the `expo` fixture

- E2E should ✅ 
- Building both fixture apps should work on all platforms (Android, iOS and Web):
  - Bare React Native:
    - `cd fixture/react-native`
    - `yarn`
    - `cd ios && bundle exec pod install && cd ..`
    - `yarn start` and then press `a` (for Android) or `i` (for iOS)
  - Expo:
    - `cd fixture/expo`
    - `yarn`
    - `yarn start` and then press `a` (for Android), `i` (for iOS) or `w` (for Web)

## Checklist


yarn.lock
```diff
@@ -25,6 +25,14 @@
     "@babel/highlight" "^7.23.4"
     chalk "^2.4.2"
 
+"@babel/code-frame@^7.24.1":
+  version "7.24.2"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.24.2.tgz#718b4b19841809a58b29b68cde80bc5e1aa6d9ae"
+  integrity sha512-y5+tLQyV8pg3fsiln67BVLD1P13Eg4lh5RW9mF0zUuvLrv9uIQ4MCL+CRT+FTsBlBjcIan6PGsLcBN0m3ClUyQ==
+  dependencies:
+    "@babel/highlight" "^7.24.2"
+    picocolors "^1.0.0"
```
gvarandas, line 34: We can probably remove this diff, no?
It doesn't seem like we had any changes to the dependency chain besides removing flipper.

PR #1112, opened 2024-03-21
Fix stale reference to onScroll and onLoad

### LINKED ISSUE #689: MasonryFlashList onScroll callback doesn't change the value according to the component scope
## Current behavior

MasonryFlashList onScroll callback doesn't change the value according to the component scope, e.g. in the snack below we have a FlashList and a MasonryFlashList, you can see that even when we update the state the logs from the MasonryFlashList prints out the initial value while with the FlashList it prints the updated value.

https://snack.expo.dev/@gkartalis/humiliated-popcorn

## Expected behavior

MasonryFlashList and FlashList onScroll callback should behave the same way in both cases

## To Reproduce

snack with the issue repro
https://snack.expo.dev/@gkartalis/humiliated-popcorn

1. press any button to set the state
2. scroll the bottom list (flash list) and see the logs - log prints updated state value
3. scroll the top list (MasonryFlashList) and see the logs - log prints initial state value

## Platform:

- [x] iOS
- [x] Android

## Environment

1.4.0 in the snack
1.4.0 in react native 0.66.5 https://github.com/artsy/eigen/pull/7693

naqvitalha:
## Description

resolves #689 

We had stable callbacks referring to props. For now we capture props in a ref which is updated everytime. Keeps the scope of the change small and fixes the problem.

Check if onLoad callback is still working when masonry sample loads.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 - Update internal dependency and fixture app to `react-native@0.72`.
   - https://github.com/Shopify/flash-list/pull/1076
+- Fix stale reference to onScroll and onLoad
+  - https://github.com/Shopify/flash-list/pull/1112
 
 ## [1.6.4] - 2024-03-18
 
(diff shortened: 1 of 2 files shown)


KingAmo:
should we publish a release for this ?

PR #1218, opened 2024-05-30
Fix RangeError if estimatedSize={0}

usrbowe:
## Description

If the `estimatedSize` is set to `0` (could be by accident or before the final value is calculated), the FlashList would throw 

```
RangeError: Array size is not a small enough positive integer.
```

Root cause code:
https://github.com/Shopify/flash-list/blob/cb9c30f1be11e85c726bb7a0ae82692f2e79bb11/src/GridLayoutProviderWithProps.ts#L128-L136

The error is due to the using `??` instead of `||`, so the `estimatedItemCount` would divide the `renderWindowSize` with `0`. Then the value of `estimatedItemCount` would be `Infinity`.

This value is later passed to new instance of `AverageWindow` as `size` as `Infinity`. 

The actual error is thrown from this line:
https://github.com/Shopify/flash-list/blob/cb9c30f1be11e85c726bb7a0ae82692f2e79bb11/src/utils/AverageWindow.ts#L4-L10

Creating `new Array(Math.max(1, Infinity))` would throw the **RangeError**.

- [ ]

## Checklist

### DIFF
--- a/src/GridLayoutProviderWithProps.ts
+++ b/src/GridLayoutProviderWithProps.ts
@@ -131,7 +131,7 @@ export default class GridLayoutProviderWithProps<T> extends GridLayoutProvider {
         (this.props.horizontal
           ? renderWindowSize.width
           : renderWindowSize.height) /
-          (this.props.estimatedItemSize ?? this.defaultEstimatedItemSize)
+          (this.props.estimatedItemSize || this.defaultEstimatedItemSize)
       )
     );
     this.averageWindow = new AverageWindow(


naqvitalha:
Hi! 👋 Thank you for your contribution to FlashList!

We're now focusing exclusively on **FlashList v2** and **React Native's New Architecture**. As part of this transition, we're closing older pull requests to better focus our review efforts on v2-compatible changes.

**If your changes are still relevant:**
- Please verify they work with FlashList v2
- Rebase against the latest main branch
- Open a new PR with v2 compatibility in mind
- We'll be happy to review it!

**Resources:**
- [Contributing Guide](https://shopify.github.io/flash-list/docs/contributing)
- [FlashList v2 Documentation](https://shopify.github.io/flash-list/)

We truly appreciate your effort and contribution. Thank you for helping make FlashList better! 🙏

PR #1236, opened 2024-06-17
Upgrade recyclerlistview to v4.2.1

### LINKED ISSUE #784: List not rendering items after initial data array is empty
How to reproduce:

1. Have an empty data set. Use EmptyListComponent
 - EmptyListComponent will render correctly.
2. Add an item to the data set without destroying the FlashList i.e via state change.
 - EmptyListComponent will disappear, but new item is not rendered.

However if the data set is not empty on initialization then everything works as expected.

So if the data array has an item it will render the item. If the item is removed then EmptyListComponent is rendered as expected.
Then If a new item is added then it will also render as expected. 

This leads me to believe that there is an issue that occurs when initializing a FlashList with an empty data set.

Might be related to: https://github.com/Shopify/flash-list/issues/757 however I wasn't using a horizontal list

naqvitalha:
## Description

Fixes #784 and potential issue where a blank space can be visible below the first row when list only has one item. This PR also adds functionality to allow access to scroll view's native ref.

Both issues were in recyclerlistview which is being update. RLV release can be found [here](https://github.com/Flipkart/recyclerlistview/releases/tag/4.2.1).

- [x] Check for regressions

## Checklist


gvarandas (commented):
Changes LGTM (thanks for fixing the e2e path)!
It looks like the tests failed on CI, though, so we might need to address those.

naqvitalha:
@gvarandas Working on it. They pass locally but fail on CI for some reason. Trying to fix artifact upload job to see what's happeing.

PR #1484, opened 2025-01-22
fix(android): support RN 0.77

huextrat:
## Description

Android is not building with RN 0.77 and FlashList

fixes: https://github.com/Shopify/flash-list/issues/1466

@tomekzaw patch seems to work well, so here is the PR

- [ ]

## Checklist


efstathiosntonas:
@huextrat is this backwards compatible with older versions of react-native? Asking because `@rnmapbox/maps` needs this fix too.

PR #1574, opened 2025-03-19
Use Github actions for publishing

naqvitalha:
## Description

Move to github actions for publishing the package.

- [ ]

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: ".nvmrc"
+          registry-url: 'https://registry.npmjs.org'
 
       - name: Install dependencies
         run: yarn install --frozen-lockfile
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,62 @@
+name: Create version packages PR or release
+
+on:
+  release:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
+name: Create version packages PR or release
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch: # Allow manual triggering of the workflow
+    inputs:
+      tag:
+        description: "Tag to release"
+        required: true
```
gvarandas, line 10: You can actually drop this if you want, as you can select the tag when running the workflow.

.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
+          # Use the tag associated with the release
+          ref: ${{ env.TAG_NAME }}
+
+      - name: Setup Node from .nvmrc
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
```
gvarandas, line 37: You also need to add the following config here:
```yaml
registry-url: 'https://registry.npmjs.org'
```
Otherwise the NPM credentials won't be correctly applied.

PR #1617, opened 2025-04-15
FlashList v2

naqvitalha:
## Description

This PR adds a new core for new arch that will replace recyclerlistview dependency.

- [ ]


fixture/react-native/index.js
```diff
@@ -2,9 +2,11 @@
  * @format
  */
 
-import { AppRegistry } from "react-native";
+import { AppRegistry, I18nManager } from "react-native";
 
 import App from "./src/App";
 import { name as appName } from "./app.json";
 
+I18nManager.forceRTL(false);
```
abhiin1947, line 10: Should we delete this?

src/recyclerview/components/StickyHeaders.tsx
```diff
@@ -0,0 +1,210 @@
...
+      newNextStickyIndex !== nextStickyIndex
+    ) {
+      setStickyIndices({
+        currentStickyIndex: newStickyIndex,
+        nextStickyIndex: newNextStickyIndex,
+      });
+    }
+  }, [currentStickyIndex, nextStickyIndex, recyclerViewManager, sortedIndices]);
+
+  useEffect(() => {
```
abhiin1947, line 103: @naqvitalha Maybe the bug we saw is because the current index is computed in a useEffect instead of a layout effect. This would be computed well after pain right?

abhiin1947 (approved):

Issue #1669, opened 2025-05-20
[V2] List items not rendering when using React.StrictMode


j0sip:
## Current behavior

When using a [`React.StrictMode`](https://react.dev/reference/react/StrictMode), FlashList does not render cells correctly. Only the initial number of cells is rendered, but after scrolling past those, no new cells appear.

Is this a known issue and is there a plan to add support for StrictMode in near future?
<br>
![Image](https://github.com/user-attachments/assets/ffec6fc5-fbff-42b6-8f95-86f87d9443c4)

## Expected behavior

Enabling React.StrictMode should not break FlashList's cell rendering.

## To Reproduce

In fixture app, wrap the content of `App` into a  `<React.StrictMode>` component:

- in `fixture/react-native/src/App.tsx`:

```
const App = () => {
  return (
    <React.StrictMode>
      <DebugContextProvider>
        <NavigationTree />
      </DebugContextProvider>
    </React.StrictMode>
  );
};
```

## Platform:

- [ x] iOS
- [ x] Android

## Environment

2.0.0-alpha.11

naqvitalha:
There might be an issue in our sample app setup. When I wrap one of the samples with StrictMode it works. Let me know if that helps. Just try wrapping FlashList with strict directly and not the app.

naqvitalha:
Thanks, I'll look into it. We do want to support the strict mode but it's not on top of the priority list. I'll get to it eventually but I'd really appreciate a PR!

j0sip:
Hey @naqvitalha , I've created a PR that fixes the issue: https://github.com/Shopify/flash-list/pull/1695

Could you take a look?

PR #1695, opened 2025-06-02
Fix incorrect rendering when StrictMode is enabled

j0sip:
## Description

Fixes (#1669)

This PR fixes the issue of flash-list not rendering items correctly when React StrictMode is on. 

When StrictMode is on, React runs setup and cleanup one extra time. Since `_isDisposed` flag in `recyclerViewManager` is not reset during setup, `RecyclerViewManager` incorrectly stays in disposed state after initialization. 

Fixes the issue by balancing the setup and cleanup calls in `useEffect` hook. 

- [ ]

After wrapping the app in StrictMode component, rendering of items now works correctly. Example in fixture app:

flash-list-fix

### DIFF
--- a/src/recyclerview/RecyclerViewManager.ts
+++ b/src/recyclerview/RecyclerViewManager.ts
@@ -304,6 +304,12 @@ export class RecyclerViewManager<T> {
     return this.updateScrollOffset(this.getAbsoluteLastScrollOffset());
   }
 
+  restoreIfNeeded() {
+    if (this._isDisposed) {
+      this._isDisposed = false;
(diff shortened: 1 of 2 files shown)


j0sip:
I have signed the CLA!

j0sip:
I have signed the CLA!

naqvitalha (approved):

naqvitalha:
Thanks for the PR, I'll pull it in once the tests pass. Since Detox doesn't support 0.79 yet so e2e will most likely fail but for now it's okay.

PR #1737, opened 2025-06-16
Fix list auto scroll to bottom with threshold -1

### LINKED ISSUE #1736: List auto scrolls with autoscrollToBottomThreshold: undefined or -1
## Current behavior

When I reload my list, the content scrolls to bottom even though I'm not setting `autoscrollToBottomThreshold`. If I set it to -1, the problem still happens, because there's a bug on:

https://github.com/Shopify/flash-list/blob/318b6e634a7217c201c4ccd9289036f59ed7a33b/src/recyclerview/hooks/useBoundDetection.ts#L96
where it doesn't check if the property is >= 0.

It is currently:
```js
if (!isHorizontal) {
```
but should be:
```js
if (!isHorizontal && autoscrollToBottomThreshold >= 0) {
```
I will open a PR for that :)

## Expected behavior

Don't auto scroll if autoscrollToBottomThreshold is not specified.

## To Reproduce

- Create a simple list with `onEndReached` or `onStartReached` (this is important, otherwise it will early return)
- Populate the list with a couple of items in a way that it doesn't fill the screen. (so it reached the bottom)
- Reset the list to `data={undefined}` (apparently this is also important)
- Add items to the list in a way that it fills more than the screen height.
- Notice that, after rendering the items, it either scrolls a few pixels down (if you have content container padding) or completely scrolls to the bottom of the list.

I'm using masonry with 2 columns, but the issue happens without masonry too.
Sometimes it scrolls only a few pixels, if you have content container padding, instead of scrolling all the way to the bottom.

## Logs

Image

This is the case where it only scrolls a few pixels 👆🏻
which I'm pretty sure is just my `contentContainerStyle` `paddingBottom: 16`

## Platform:

- [x] iOS (very rare, but I've seen it happen)
- [x] Android

The difference between platforms might be because I have quite slow Android but a good iPhone.

## Environment

2.0.0-rc.8

pakerwreah:
## Description
When I reload my list, the content scrolls to bottom even though I'm not setting `autoscrollToBottomThreshold`. If I set it to -1, the problem still happens, because it's the same value the code coalesces to for undefined.

Fixes https://github.com/Shopify/flash-list/issues/1736

- [ ]

### DIFF
--- a/src/recyclerview/hooks/useBoundDetection.ts
+++ b/src/recyclerview/hooks/useBoundDetection.ts
@@ -94,7 +94,7 @@ export function useBoundDetection<T>(
       }
 
       // Handle auto-scrolling to bottom for vertical lists
-      if (!isHorizontal) {
+      if (!isHorizontal && autoscrollToBottomThreshold >= 0) {
         const autoscrollToBottomThresholdDistance =
           autoscrollToBottomThreshold * visibleLength;
 


pakerwreah:
I have signed the CLA!

naqvitalha (approved):
Thanks for fixing this!

PR #1842, opened 2025-08-21
Fix Jest mock returning undefined in v2.0.0

### LINKED ISSUE #1834: Jest mock returns `undefined`
## Current behavior

I'm trying to test my component which is based on `FlashList` (right now 2.0.0 as I'm migrating) and the mock returns `undefined` instead of mocked component.

## Expected behavior
The mock is successfully returned.

## To Reproduce
Just import flashlist mock in your `jestSetup`

## Platform:

- [ ] iOS
- [ ] Android

## Environment
2.0.0, on the 1st version it works fine

naqvitalha:
## Description

This PR fixes the issue where the Jest mock for FlashList returns `undefined` in version 2.0.0. Since recyclerview is exported as default, the existing mock is no longer needed.

Fixes #1834

### DIFF
--- a/src/__tests__/LayoutCommitObserver.test.tsx
+++ b/src/__tests__/LayoutCommitObserver.test.tsx
@@ -53,8 +53,11 @@ describe("LayoutCommitObserver", () => {
       />
     );
 
-    render(content);
+    const renderResult = render(content);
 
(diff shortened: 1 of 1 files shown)


glad-shopify (approved):
Changes LGTM!

PR #1870, opened 2025-08-28
Apply initialScrollOffset to it initial scroll position

joluet:
## Description

Fixes https://github.com/Shopify/flash-list/discussions/700

Adds a new property `initialScrollOffset` which allows applying an offset to `initialScrollIndex`.

### DIFF
--- a/src/FlashListProps.ts
+++ b/src/FlashListProps.ts
@@ -167,6 +167,12 @@ export interface FlashListProps<TItem>
   initialScrollIndex?: number | null | undefined;
 
   /**
+   * Offset to apply to initial scroll position as defined by initialScrollIndex.
+   * Ignored if initialScrollOffset is not set.
+   */
+  initialScrollOffset?: { x: number; y: number } | null | undefined;
+
+  /**
    * Used to extract a unique key for a given item at the specified index.
    * Key is used for optimizing performance. Defining `keyExtractor` is also necessary
    * when doing [layout animations](https://shopify.github.io/flash-list/docs/guides/layout-animation)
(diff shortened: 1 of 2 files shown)


joluet:
I have signed the CLA!

naqvitalha:
I like this feature but I think it should be called `initialScrollIndexParams` and we can reuse the contract from `scrollToIndex`: https://github.com/Shopify/flash-list/blob/47db753e4e2a83fa2eba583bed254314a5790076/src/FlashListRef.ts#L17.
`initialScrollOffset` sounds like something that should work on it's own.

documentation/docs/fundamentals/usage.md
```diff
@@ -223,6 +222,15 @@ Instead of starting at the top with the first item, start at `initialScrollIndex
 initialScrollIndex?: number;
 ```
 
+### `initialScrollIndexParams`
+
+Additional configuration for initialScrollIndex. Use viewOffset to apply an offset to the initial scroll position as defined by initialScrollIndex.
+Ignored if initialScrollIndex is not set.
+
+```tsx
+initialScrollIndexParams?: { viewOffset?: { x: number; y: number } };
```
naqvitalha, line 231: viewOffset can just be a number since we'll already know if FlashList is horizontal or not. Can you make this change and we can pull it in?

naqvitalha (changes requested):

naqvitalha (approved):

khushal87:
Hey @naqvitalha @joluet, can we also set viewPosition and animated somehow? If I would like re-center my view after scroll to first index? Does it make sense to you? I can work on it if it does.

PR #1910, opened 2025-09-16
Fix crash in scrollToEnd

robingullo:
## Description

In rare cases where the list unmounts just after a scrollToEnd, you get an unhandled `TypeError: Cannot read property 'scrollToEnd' of null` coming from this line:
```
setTimeout(function () {
    scrollViewRef.current.scrollToEnd({ animated: animated });
}, 0);
```
It might be a race condition with the `setTimeout`.

Replacing `current!.scrollToEnd` with `current?.scrollToEnd` has no other impact since it's a side effect.

### DIFF
--- a/src/recyclerview/hooks/useRecyclerViewController.tsx
+++ b/src/recyclerview/hooks/useRecyclerViewController.tsx
@@ -289,7 +289,7 @@ export function useRecyclerViewController<T>(
           }
         }
         setTimeout(() => {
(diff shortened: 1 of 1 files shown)


naqvitalha:
@robingullo Thanks for catching this! Can you sign the cla? We can then pull this in.

naqvitalha:
I'll be closing this due to missing cla and lack of response. I'll raise a PR separately as this is worth fixing.

Issue #1915, opened 2025-09-21
Scrolling with startRenderingFromBottom is WAY slower


billnye2:
## Description

The only thing I am changing is whether startRenderingFromBottom is false or true. When false, scrolling is fast. When true, scrolling is slow in the sense that I am swiping as fast as I can and the view is just scrolling very slowly. I notice that once I scroll past say a certain item, if I reverse the scroll direction then it's back to really fast while scrolling in that opposite direction. So like, it's only fast in the zone that I have already visited. Once I scroll back to items I haven't seen yet, then it slows right back down.

## Platform

Android

</details>

**FlashList version:** 2.0.3

billnye2:
I've discovered that it is much faster to just use the old inverted hack by rotating the list and rotating each item. The list is way faster this way

joaqo:
Having the same issue.

Issue #1872, opened 2025-08-29
Flashlist:v2.0.3: Type chat render from bottom not working.


VincentLe94:
Hi team, I am migrated from v1 to v2.0.3 to support new arch. But flag startRenderingFromBottom not working expected.

<FlashList
              ref={flashtListRef}
              data={dataMessage ?? []}
              renderItem={renderItem}
              keyExtractor={(item, index) =>
                `${item.id}-item-message`
              }
              onScroll={handleScroll}
              maintainVisibleContentPosition={{
                autoscrollToBottomThreshold: 0.2,
                startRenderingFromBottom: true,
              }}
              showsVerticalScrollIndicator={false}
            />

LuongTruong:
@VincentLe94 I am wondering if your are facing the same issue as me. The list is simply render from bottom, but not REVERT the items in the list. 🤔

flessard:
@naqvitalha Please roll back to the old mechanic with invert; the new one is really buggy and not smooth.

evelant:
Hmm it looks like there may be a bug with `onStartReached` -- it works when scrolling up but only when you reach the very top of the list. If you're near the top and scroll _down_ it starts loading more at the top if within the threshold. This is on 2.1.0. It seems like there's a logic bug with `onStartReached` where it's listening for scroll down events rather than scroll up.

In addition to this the transition to 2.0 is a bit confusing with regard to inverted lists per the docs. You need to do more than the docs mention. Here's how it plays out:

* Reverse the sort of your data from how it was using `inverted`
* Replace `onEndReached` with `onStartReached`
* Switch `ListHeaderComponent` with `ListFooterComponent`
* Remove `inverted` prop, replace with `maintainVisibleContentPosition` with `startRenderingFromBottom: true`

The key being the list is no longer inverted. It's in "normal" top to bottom order except scrolling starts at the bottom. The `onStartReached` behavior does seem like a genuine bug however.

minhna:
I tried to set `onEndReachedThreshod` value to `0`, when I scroll up to the top of the list, it loads more items but render them down below. In other words, the list stays at the top. It's not expected behavior.

LuongTruong:
Hello guys, thanks for the input. After reading the document once again, I see that the list need to be reversed by ourselves before using. However, the thing is replacing `onEndReached` by `onStartReached`, I saw there are some tickets saying it is not working well.

I am wondering if the version `2.1.0` fixes the issue?

[...]

Issue #1864, opened 2025-08-26
Glith on rendering new item


umardev500:
## Current behavior

Currently when I added item to top then never rendered first time that with glitch on scrolling

## Expected behavior

smooth scrolling while rendering new item never rendered before

## To Reproduce

Already add keyExtractor and more

## Platform:

- [x] iOS
- [ ] Android

## Environment

```json
 "dependencies": {
    "@lodev09/react-native-true-sheet": "^2.0.6",
    "@react-native-vector-icons/icomoon": "^12.2.0",
    "@react-native/new-app-screen": "0.80.2",
    "@react-navigation/bottom-tabs": "^7.4.6",
    "@react-navigation/elements": "^2.6.3",
    "@react-navigation/native": "^7.1.17",
    "@react-navigation/native-stack": "^7.3.25",
    "@shopify/flash-list": "^2.0.3",
    "clsx": "^2.1.1",
    "formik": "^2.4.6",
    "nativewind": "^4.1.23",
    "react": "19.1.0",
    "react-native": "0.80.2",
    "react-native-edge-to-edge": "^1.6.2",
    "react-native-flash-message": "^0.4.2",
    "react-native-image-picker": "^8.2.1",
    "react-native-keyboard-controller": "^1.18.5",
    "react-native-reanimated": "^3",
    "react-native-safe-area-context": "^5.6.1",
    "react-native-screens": "^4.15.2",
    "yup": "^1.7.0"
  },
```

### List

```tsx
import { ChatListing } from "@/src/features/chat/components/molecule";
import { ChatMessage } from "@/src/features/chat/types";
import { faker } from "@faker-js/faker";
import { FlashList, FlashListRef, ListRenderItem } from "@shopify/flash-list";
import React, { useState } from "react";
import { useSafeAreaInsets } from "react-native-safe-area-context";

// Generate initial 50 dummy items
const generateItems = (start: number, count: number) => {
  const timestamp = faker.date.recent({ days: 7 });

  return Array.from({ length: count }, (_, i) => ({
    id: String(start + i),
    text: faker.lorem.sentence(),
    sender: Math.random() > 0.5 ? "u1" : "u2",
    timestamp: timestamp.toISOString(),
  }));
};

const CURRENT_USER = "u1";

interface Props {
  ref?: React.RefObject<FlashListRef<ChatMessage> | null>;
}

export const ChatList: React.FC<Props> = ({ ref }) => {
[...]

Issue #1844, opened 2025-08-22
FlashList v2 Chat UX Regression: Inconsistent scroll behavior with paginated messages


EL-BADI:
## Summary
After migrating from FlashList v1 to v2, chat interfaces with paginated message loading experience a confusing UX regression where the scroll direction for viewing messages and loading more messages are inconsistent.

## Expected Behavior (v1 with `inverted={true}`)
- ✅ Latest messages appear at the bottom
- ✅ Users scroll **up** to see older messages  
- ✅ Users scroll **up** to trigger `onEndReached` and load more messages
- ✅ Consistent scroll direction for both viewing and loading

## Current Behavior (v2 with `maintainVisibleContentPosition`)
- ❌ Latest messages appear at the top
- ❌ Users scroll **down** to see older messages
- ❌ Users must scroll **up** to trigger `onStartReached` and load more messages
- ❌ **Inconsistent scroll direction**: down to view, up to load more

## Context
This issue occurs in a chat application where messages are loaded via **paginated Convex queries**. The typical flow is:

1. Initial load: Fetch most recent N messages
2. User scrolls to view older messages  
3. Trigger pagination to load next batch of older messages
4. New messages arrive in real-time and are appended

## Problem Description

In v2, the deprecation of `inverted` prop forces developers to use `maintainVisibleContentPosition` with `startRenderingFromBottom: true`. However, this creates a fundamental UX issue:

1. **Visual Layout**: Messages appear with newest at top (like Twitter/social feeds)
2. **Loading Pattern**: Must scroll to opposite direction to load more content
3. **User Confusion**: Users expect to scroll in one direction for both viewing and loading

### v1 Implementation (Working UX with Convex pagination)
```tsx
<FlashList
  data={messages} // From usePaginatedQuery(api.messages.list)
  renderItem={renderMessage}
  inverted={true}
  onEndReached={() => {
    // Load more older messages via Convex pagination
    loadMore();
  }}
  onEndReachedThreshold={0.5}
/>
```

### v2 Implementation (Broken UX with Convex pagination)
```tsx
<FlashList
  data={messages} // From usePaginatedQuery(api.messages.list)
  renderItem={renderMessage}
  maintainVisibleContentPosition={{
    startRenderingFromBottom: true,
    autoscrollToBottomThreshold: 0.2,
  }}
  onStartReached={() => {
    // Load more older messages - but user scrolled DOWN to see old messages!
    loadMore(); // This creates cognitive dissonance
  }}
[...]

PR #1932, opened 2025-10-03
Fix onStartReached detection

evelant:
## Description

When scrolling up with `onStartReached` `useBoundDetection` does not properly check scroll position while scrolling up. This resulted in `onStartReached` only firing when reaching the very top of the list and also firing while near top of the list and scrolling down. It was missing `firstItemOffset` in the calculation and missing `pendingStartReached.current = false` on data changes.

Fixes (issue #) https://github.com/Shopify/flash-list/issues/1872

Likely fixes 

[#1915
](https://github.com/Shopify/flash-list/issues/1915)
https://github.com/Shopify/flash-list/issues/1844
https://github.com/Shopify/flash-list/issues/1864

- [ ]


evelant:
I signed the cla

PR #1952, opened 2025-10-14
Fix sticky header apprearance in flashlist v2

phamxdinhhien:
### Description
Fixes an issue where sticky headers become active before the header item actually reaches the top of the list when `firstItemOffset` is applied to the RecyclerView.

### Problem
When using `stickyHeaderIndices` with a FlashList that has content insets, padding, or safe area adjustments (represented by `firstItemOffset`), the sticky header would appear too early—before the actual header item reached the top of the visible area. This created a jarring user experience where the sticky header appeared while the original header was still visible below it.

### Root Cause
The position calculations in `StickyHeaders.tsx` were not consistently handling `firstItemOffset`:
1. The binary search for finding the current sticky index was comparing scroll offset against absolute item Y positions without normalizing for `firstItemOffset`
2. The next sticky header Y position calculation was adding `firstItemOffset`, causing it to be double-counted

This resulted in coordinate space misalignment, leading to premature sticky header activation.

### Solution
This PR ensures all position comparisons happen in the same coordinate space by:

1. **Normalizing item positions in binary search** (line 95):
   - Subtract `firstItemOffset` from item Y positions when comparing against scroll offset
   - This ensures the comparison accounts for the content inset

2. **Removing double-counting in next sticky calculation** (line 109):
   - Remove the addition of `firstItemOffset` since it's already included in the layout Y position
   - This prevents the next sticky header from being calculated with an incorrect offset

### Changes
- Modified `src/recyclerview/components/StickyHeaders.tsx`

### Testing
- [x] Tested with FlashList using `contentContainerStyle` with `paddingTop`
- [x] Tested with `contentInset` on iOS
- [x] Verified sticky headers now activate precisely when items reach the top
- [x] Confirmed smooth push/pull animation between consecutive sticky headers

### Impact
- Fixes sticky header timing issues when `firstItemOffset` > 0
- No breaking changes—purely a bug fix
- Improves UX for lists with content insets or safe area adjustments


naqvitalha:
Can you provide an expo snack showcasing the problem? I check our contact sample and it works just fine with padding applied.

focux:
> Can you provide an expo snack showcasing the problem? I check our contact sample and it works just fine with padding applied.

I'm not OP but I think there's a repro here https://github.com/Shopify/flash-list/issues/1959
I'm experiencing this issue as well.

RodolfoSilva:
@naqvitalha here’s the Snack https://snack.expo.dev/@rodolfosilva/laughing-red-watermelon?platform=ios

This issue only occurs on native, it works fine on the web.

RodolfoSilva:
@naqvitalha another issue related to this is that the header becomes sticky too often.

RodolfoSilva:
I've applied this patch, but it doesn't fix the issue.

[@shopify+flash-list+2.2.0.patch](https://github.com/user-attachments/files/23279450/%40shopify%2Bflash-list%2B2.2.0.patch)

naqvitalha:
Closing — this is superseded by PR #2105 (merged), which fixed the root cause by correcting `measureParentSize` to exclude x/y offsets on Fabric. The sticky header offset issue described here is resolved on `main`. Thanks for the contribution!

PR #2053, opened 2026-01-06
Ensure average measurement is always positive

mhoran:
The calculation of the next average might be negative due to floating point precision errors. This will then compound over time and create layout issues.

## Description

If a list contains a mix of hidden and visible items with vastly different sizes, it's easy to end up with a moving average that ends up negative. Over time this compounds, eventually causing major layout issues.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 env:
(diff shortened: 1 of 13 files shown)


mhoran:
I have signed the CLA!

naqvitalha:
Thanks for the PR, looks good.

alizahid:
Does this fix help with the screenshot or is that a separate issue?

Screenshot 2026-03-05 at 5 11 39 AM

PR #2054, opened 2026-01-10
Improve migration by including find&replace regex for deprecated props

audn:
Added regex instructions for removing deprecated props.

## Description

Fixes (issue #)

- [ ]

### DIFF
--- a/documentation/docs/v2-migration.md
+++ b/documentation/docs/v2-migration.md
@@ -43,8 +43,6 @@ The following props have been deprecated and should be removed from your FlashLi
 />
 ```
 
-##### 
-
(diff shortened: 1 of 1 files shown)


naqvitalha:
Thanks for the contribution! We're going to pass on this one though — the regex approach is a bit fragile since it won't handle cases like nested braces (e.g., `estimatedListSize={{ height: 100, width: 200 }}`) or multi-line prop values, and we'd rather not add something to the migration guide that might give users a false sense of completeness. The list of deprecated props is short enough that removing them manually is straightforward. Appreciate you thinking about ways to make the migration easier!

PR #2096, opened 2026-02-11
fix(layout): items height normalization after layout change

### LINKED ISSUE #881: Flashlist horizontal wrong height item
## Current behavior
Hi, first of all, you lib are cool, but I'm facing with issue, when I try use it in horizontal mode. Item of list are cutting in bottom
Снимок экрана 2023-07-24 в 10 29 33

https://github.com/Shopify/flash-list/assets/23406402/30ef17a5-28fd-4dab-90a7-9e2c574deb8d

code:
```ts
export const { width: WIDTH, height: HEIGHT } = Dimensions.get('window');

cont Comp = (props) => {
if({!data?.length) return <></>

return (
<FlashList
      ref={ref}
     data={data}
     renderItem=(({ item }) => <Card width={cardWidth} {...item} />)
      renderScrollComponent={ScrollView}
      bounces
     estimatedItemSize={WIDTH}
      showsVerticalScrollIndicator={false}
      showsHorizontalScrollIndicator={false}
      scrollEventThrottle={32}
      onEndReachedThreshold={2}
      {...props}
    />
)
}

```

## Expected behavior

Correct item size without cutting

## To Reproduce
Try to add different height for items

## Platform:

- [x ] iOS
- [ x] Android

## Environment
 "react-native": "0.71.8",
"@shopify/flash-list": "^1.4.2"

VickyStash:
## Description

Fixes https://github.com/Shopify/flash-list/issues/881

This PR fixes several bugs in `LinearLayoutManager.normalizeLayoutHeights` that caused horizontal FlashList items to render with inconsistent heights after a layout change, item deletion, or item insertion.  

### Reproduction snack 1: https://snack.expo.dev/@vikstash/shallow-orange-candies

**Steps:**
1. Initially, all items have a max item height (all the same).
2. Tap the button to increase the item width (it makes the max item height decrease)
3. Scroll through items and see that their height isn't normalized anymore.

<details><summary>video</summary>

https://github.com/user-attachments/assets/fdc516b8-1cee-4c52-859f-af03b6fa6d1a
</details>

### Reproduction snack 2: https://snack.expo.dev/@vikstash/blissful-orange-apples

**Steps:**
1. Initially, all items have a max item height (all the same).
2. Add/remove the highest items several times in a row.
3. Scroll through items and see that their height isn't normalized anymore.

<details><summary>video</summary>

https://github.com/user-attachments/assets/9612af4f-451e-4d15-948c-bbcb744dd697
</details>

                                                                                                                                                                         
  ### Problem                                                                                                                                                            
 ### 1. Shrink path leaves items stuck at `minHeight: 0`
When the tallest item's measured height decreases (e.g., after a screen resize), the method [correctly sets ](https://github.com/Shopify/flash-list/blob/5a15a807a77f8b3cf17e9f3efccaf9d82e703dab/src/recyclerview/layout-managers/LinearLayoutManager.ts#L126-L129)`targetMinHeight = 0` and requests a repaint so items can be re-measured. 
However, it also [updates](https://github.com/Shopify/flash-list/blob/5a15a807a77f8b3cf17e9f3efccaf9d82e703dab/src/recyclerview/layout-managers/LinearLayoutManager.ts#L139) `this.tallestItemHeight` to the new (shorter) value on the same cycle.
On the next cycle after repaint, the [guard](https://github.com/Shopify/flash-list/blob/5a15a807a77f8b3cf17e9f3efccaf9d82e703dab/src/recyclerview/layout-managers/LinearLayoutManager.ts#L124) `newTallestItem.height !== this.tallestItemHeight` evaluates to `false` (the value was already stored), so the entire normalization block is skipped. 
All items remain with `minHeight: 0` permanently.

### 2. Deleting the tallest item leaves stale tracking
Deleting the tallest item leaves stale tracking. When the item tracked as `tallestItem` is removed from the list, the                                                                                                
manager keeps the stale reference and stale `tallestItemHeight`. The shrink/grow guard then never triggers, and the remaining items stay sized to a height that no longer exists in the data.

### 3. Newly added items aren't normalized to the current tallest   
When new items are inserted and their measured height is less than or equal to the current `tallestItemHeight`, the existing block is skipped (no new tallest detected), so the added items render at their natural height instead of being padded to match siblings. 

  ### Fix

  - **Shrink path:** reset `this.tallestItem`/`this.tallestItemHeight` when `targetMinHeight === 0` so the next cycle re-detects the tallest item and re-applies `minHeight` to all layouts.                                                                                                                        
  - **Tallest item removed:** if the tracked `tallestItem` is no longer in `this.layouts`, clear `minHeight` on remaining layouts, reset tracking, and request a repaint so heights get re-measured.                                                                                                                 
  - **New items added:** after the shrink/grow branch, normalize any layout whose `minHeight` doesn't match the current `tallestItemHeight` so freshly inserted items inherit the shared height.       

Focus on the conditional reset logic in `normalizeLayoutHeights` 
- [ ] verify that resetting `tallestItem`/`tallestItemHeight` in the shrink path doesn't break the grow
  path (when a new item is taller than the current tallest).
- [ ] verify the early-return on deletion doesn't double-fire, and the added-items normalization doesn't loop over large lists every measurement


VickyStash:
I have signed the CLA!

VickyStash:
Hey @naqvitalha, what do you think?

PR #2104, opened 2026-02-16
Fix disappearing cards in masonry when optimizeItemArrangement is dis…

### LINKED ISSUE #2103: Disappearing items in Masonry layout
## Description

When using FlashList with masonry enabled and `optimizeItemArrangement` set to false, items can sometimes disappear during scrolling. The issue is more likely to happen where there is a difference in column heights.

## Current behavior

When there is a significant discrepancy of visible items' indices, cards sometimes disappear unexpectedly  while scrolling. The root cause seems to be that the layout uses binary search for calculating visible items; and because binary search expects sorted items, the algorithm fails when there is a difference in indices between columns 

Video: 

https://github.com/user-attachments/assets/68e9a65a-4981-4675-b705-4985e300f491

## Expected behavior

Cards should not disappear or flicker as user scrolls.

## Reproduction

The issue can be reproduced in fixture app.
1. Disable `optimizeItemArrangement` in [ComplexMasonry](https://github.com/Shopify/flash-list/blob/main/fixture/react-native/src/ComplexMasonry.tsx#L189) example:
```
        optimizeItemArrangement={false}
```

2. Open the ComplexMasonry screen and scroll down (see video above)

## Platform

- [x] iOS
- [x] Android
- [ ] Web (if applicable)

## Environment

</details>

**FlashList version:** `"2.2.2"`

## Checklist

j0sip:
## Description

Fixes [(issue #2103)](https://github.com/Shopify/flash-list/issues/2103)

This PR fixes an issue in masonry layout where cards unexpectedly disappear as user scrolls when `optimizeItemArrangement` is disabled. 

It adds a new arrays of sorted layout indices to MasonryLayoutManager, which enables correct binary search when calculating visible items. Current implementation of binary search fails when `optimizeItemArrangement` is disabled, because the array of layouts is not sorted by y position. This leads to incorrect results when column heights are different.

The fix only affects masonry layout and only when  `optimizeItemArrangement === false`, since the issue is not occurring when the prop is enabled.

- [ ]

Video (after disabling optimizeItemArrangement in ComplexMasonry from fixture app:

https://github.com/user-attachments/assets/e6bde5fd-dbb3-4954-b4a0-8af448134e26

### DIFF
--- a/src/__tests__/MasonryLayoutManager.test.ts
+++ b/src/__tests__/MasonryLayoutManager.test.ts
@@ -199,4 +199,56 @@ describe("MasonryLayoutManager", () => {
       expect(getAllLayouts(manager).length).toBe(0);
     });
   });
+
+  describe("Visibility with uneven column heights (optimizeItemArrangement: false)", () => {
+    const unevenParams = {
+      windowSize,
+      maxColumns: 2,
+      optimizeItemArrangement: false,
+    };
+
+    // Creates a masonry layout with different column heights.
+    const createUnevenMasonryLayout = () => {
+      const manager = createLayoutManager(
+        LayoutManagerType.MASONRY,
+        unevenParams
+      );
+      const layoutInfos = [
+        createMockLayoutInfo(0, 200, 100),
+        createMockLayoutInfo(1, 200, 50),
+        createMockLayoutInfo(2, 200, 100),
+        createMockLayoutInfo(3, 200, 50),
+        createMockLayoutInfo(4, 200, 100),
+        createMockLayoutInfo(5, 200, 50),
+        createMockLayoutInfo(6, 200, 100),
+        createMockLayoutInfo(7, 200, 50),
+      ];
+      manager.modifyLayout(layoutInfos, 8);
+      return manager;
+    };
+
+    it("should find items visible at the top of the layout", () => {
+      const manager = createUnevenMasonryLayout();
+
+      const visible = manager.getVisibleLayouts(0, 100);
+
+      expect(visible.includes(0)).toBe(true);
+      expect(visible.includes(1)).toBe(true);
+      expect(visible.includes(3)).toBe(true);
+      expect(visible.includes(6)).toBe(false);
+    });
+
+    it("should find visible items when viewport is in the middle (column heights diverge)", () => {
+      const manager = createUnevenMasonryLayout();
+
+      const visible = manager.getVisibleLayouts(100, 200);
+
+      expect(visible.includes(2)).toBe(true);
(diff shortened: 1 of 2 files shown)


naqvitalha:
Thanks for the PR! Binary search on the flat layouts array breaks in masonry with sequential placement because items zigzag between columns (e.g., y-values go [0, 0, 100, 50, 200, 100]), so binary search can miss visible items.

However, the sorting approach adds significant overhead:
- O(N log N) sort on every layout change (height measurements, new items)
- O(N) extra memory for sorted copies
- Rebuild on every `processLayoutInfo` and `recomputeLayouts` call

This is avoidable because items within each column are naturally sorted by y. With sequential placement, item `i` belongs to column `i % maxColumns`. So you can binary search per column instead of sorting the whole array:

```ts
getVisibleLayouts(start: number, end: number): ConsecutiveNumbers {
  if (this.optimizeItemArrangement) {
    return super.getVisibleLayouts(start, end);
  }

  let minIdx = Infinity, maxIdx = -1;

  for (let col = 0; col < this.maxColumns; col++) {
    // Items in column: col, col+maxColumns, col+2*maxColumns, ...
    // Their y-values are sorted — binary search works directly
    const first = this.binarySearchColumn(col, start, true);
    const last = this.binarySearchColumn(col, end, false);
    if (first !== -1) minIdx = Math.min(minIdx, first);
    if (last !== -1) maxIdx = Math.max(maxIdx, last);
  }

  return minIdx > maxIdx
    ? ConsecutiveNumbers.EMPTY
    : new ConsecutiveNumbers(minIdx, maxIdx);
}

private binarySearchColumn(
  col: number, threshold: number, findFirst: boolean
): number {
  // Items in this column at original indices: col, col+mc, col+2*mc, ...
  const mc = this.maxColumns;
  const count = Math.ceil((this.layouts.length - col) / mc);
  let left = 0, right = count - 1, result = -1;

  while (left <= right) {
    const mid = (left + right) >> 1;
    const idx = col + mid * mc;
    const layout = this.layouts[idx];
    const pos = this.horizontal ? layout.x : layout.y;
    const size = this.horizontal ? layout.width : layout.height;

    if (findFirst) {
      if (pos + size > threshold) { result = idx; right = mid - 1; }
      else { left = mid + 1; }
    } else {
      if (pos <= threshold) { result = idx; left = mid + 1; }
      else { right = mid - 1; }
    }
  }
  return result;
}
```

nameishuy:
Hi, I stuck with this issue and patch package these codes in. But it still disappearing, I tried console log in getVisibleLayouts but they didnt trigger

Issue #2107, opened 2026-02-20
Masonry items jump between columns upon size change


j0sip:
## Current behavior

In a Masonry layout with `optimizedItemArangement` enabled, items sometimes jump to other column when their height is modified. This results in bad UX.

See video below:

https://github.com/user-attachments/assets/466e1bd5-9173-4339-8405-c25ac7ac1433

## Expected behavior

There should at least be a prop that disallow items to switch columns upon size change or it should be enabled by default, at least for the visible items. 
Unexpectedly switching columns seems glitchy and leads to bad UX.

## Reproduction

Reproducible in flash-list fixture app:

1. Open ComplexMasonry screen example
2. Click on Item 9 or Item 13 to increase their size and notice how other cards jump to different column

## Platform

- [x] iOS
- [x] Android

## Environment

**FlashList version:** 2.2.2

## Checklist

- [x] I've searched existing issues and couldn't find a duplicate
- [x] I've provided a minimal reproduction (Expo Snack preferred)
- [x] I'm using the latest version of @shopify/flash-list
- [x] I've included all required information above

j0sip:
In our case, this issue prevents us from using `optimizedItemArrangement` because our list contains multiple items whose height can change dynamically and item jumping between columns is not acceptable. 

I have opened an PR (https://github.com/Shopify/flash-list/pull/2108) with a potential fix. I'm happy to make any necessary adjustments. Looking forward to the feedback.

PR #2108, opened 2026-02-20
Fix item jumps in masonry layout

j0sip:
## Description

Fixes https://github.com/Shopify/flash-list/issues/2107

This PR prevents items from "jumping" between columns in the Masonry Layout when `optimizedItemArrangement` is enabled. It ensures that resizing an item doesn't disrupt the user's visual flow by locking the column positions.

Currently, when an item’s size changes, the Masonry algorithm recomputes the entire layout for all subsequent items. While this optimization helps fill gaps, it causes visible items to jump to different columns, causing a bad UX for end users.

To improve the UX,  I've introduced a `lastLockedItemIndex` which is the index of last engaged item. It ensures that all items up to that index retain their previously calculated column. Recomputation of columns would now only be allowed for items above that index (i.e. non-visible items), so the end user wouldn't notice it.

- [ ]

Tested in ComplexMasory in fixture app:
after

### DIFF
--- a/src/recyclerview/RecyclerViewManager.ts
+++ b/src/recyclerview/RecyclerViewManager.ts
@@ -119,6 +119,7 @@ export class RecyclerViewManager<T> {
 
       if (engagedIndices) {
         this.updateRenderStack(engagedIndices);
+        this.layoutManager.onEngagedIndicesChanged(engagedIndices);
         return engagedIndices;
       }
     }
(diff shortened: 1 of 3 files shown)


naqvitalha:
Thanks for the PR and the detailed write-up!

I think there's a simpler path here. The jumping happens because `optimizeItemArrangement` recomputes column assignments to balance column heights — that's its job. When an item resizes, it re-runs that optimization and items move to different columns.

Column locking fights against that optimization. And with multi-span items, locking can create visible gaps (e.g., a span=2 item must start at the tallest column's height, leaving empty space in shorter columns).

The simpler solution: if your items resize dynamically, use `optimizeItemArrangement: false`. Sequential placement means items never change columns — there's nothing to "jump." The layout won't be perfectly balanced, but items stay stable, which is what you want for dynamic content.

Think of it as:
- **`optimizeItemArrangement: true`** — best for items with fixed/stable heights. Produces a balanced layout.
- **`optimizeItemArrangement: false`** — best for items that resize after mounting. Items stay in their column but devs manager the sequence.

j0sip:
> I think there's a simpler path here. The jumping happens because `optimizeItemArrangement` recomputes column assignments to balance column heights — that's its job. When an item resizes, it re-runs that optimization and items move to different columns.
> 
> The simpler solution: if your items resize dynamically, use `optimizeItemArrangement: false`. Sequential placement means items never change columns — there's nothing to "jump." The layout won't be perfectly balanced, but items stay stable, which is what you want for dynamic content.


Thanks for the reply @naqvitalha . We have tried using `optimizeItemArrangement: false`, but it easily turns into unbalanced column heights for larger lists, which does not look good from UI/UX perspective. Ideally, we would want both: balanced columns and allow dynamic resizing, but I understand that's not possible right now.

Perhaps another solution that could be considered is if we extend `optimizeItemArrangement` prop to be an enum instead, something like:
- **`optimizeItemArrangement: always`** - same as current `true`
- **`optimizeItemArrangement: never`** - same as current `false`
- **`optimizeItemArrangement: onFirstLayout`** - this option can allow locked columns after first layout, preventing the jumping cards

Boolean values could still be supported for backward compatibility. Could you consider that approach?


>And with multi-span items, locking can create visible gaps (e.g., a span=2 item must start at the tallest column's height, leaving empty space in shorter columns).
> 

Gaps with multi-span appear even in current implementation when items resize, because most of the time none of the items can fit the gap. It's evident in fixture app (ComplexMasonry) as well. Additionally, I believe small gap is better than items jumping between columns.

PR #2131, opened 2026-03-03
feat(scroll): add inverted prop

### LINKED ISSUE #1844: FlashList v2 Chat UX Regression: Inconsistent scroll behavior with paginated messages

naqvitalha:
## Description

Re-introduces the `inverted` prop for FlashList v2, addressing the most requested feature gap since the v1-to-v2 migration. The implementation uses platform-specific CSS transforms — `scaleY(-1)` on iOS/web and `rotate(180deg)` on Android (for performance, see #751) — applied at the container, cell, header, and footer levels.

**Note:** On Android, the rotate optimization causes the scrollbar to appear on the left side.

Fixes #1844

- Transform cascade: container inverts the list, then each cell/header/footer counter-inverts to render content right-side-up
- Android uses `rotate(180deg)` instead of `scaleY(-1)` due to known perf issues (#751)
- Verify `onEndReached` still fires correctly in the inverted direction
- Check interaction with `maintainVisibleContentPosition` (these are independent approaches — inverted doesn't use it)

## Test plan
- [x] Unit tests pass (`yarn test` — 183 passing)
- [x] Type check passes (`yarn type-check`)
- [x] Lint passes (`yarn lint`)
- [x] Verified on iOS simulator — scrolling, loading older messages, adding new messages all work correctly
- [ ] E2e snapshot tests added (`ChatInverted.test.e2e.ts`) — reference screenshots need to be generated on first run
- [ ] Verify on Android

### DIFF
--- a/fixture/react-native/e2e/tests/Twitter.test.e2e.ts
+++ b/fixture/react-native/e2e/tests/Twitter.test.e2e.ts
@@ -17,6 +17,7 @@ describe("Twitter", () => {
   });
 
   it("with FlashList looks the same", async () => {
+    await element(by.id("ExamplesFlatList")).swipe("up", "slow", 0.3);
     await element(by.id("Twitter Timeline")).tap();
 
     const testRunScreenshotPath = await element(
@@ -51,6 +52,7 @@ describe("Twitter", () => {
     const testName = "Twitter_looks_the_same_after_orientation_change";
 
     // Go to Twitter with FlashList screen
+    await element(by.id("ExamplesFlatList")).swipe("up", "slow", 0.3);
     await element(by.id("Twitter Timeline")).tap();
     // Scroll 500px down and change orientation to lansdsape
     await scrollAndRotate("FlashList");
@@ -65,6 +67,7 @@ describe("Twitter", () => {
 
   it("is updated after refreshed", async () => {
     const testName = "Twitter_is_updated_after_refreshed";
+    await element(by.id("ExamplesFlatList")).swipe("up", "slow", 0.3);
     await element(by.id("Twitter Timeline")).tap();
 
     const flashList = element(by.id("FlashList"));
@@ -82,6 +85,7 @@ describe("Twitter", () => {
     const testName =
       "Twitter_loads_a_new_page_when_gets_to_the_bottom_of_the_list";
     await enableDebugOption(DebugOption.PagingEnabled);
+    await element(by.id("ExamplesFlatList")).swipe("up", "slow", 0.3);
     await element(by.id("Twitter Timeline")).tap();
 
     const flashList = element(by.id("FlashList"));


src/native/config/PlatformHelper.web.ts
```diff
@@ -3,6 +3,8 @@ const PlatformConfig = {
   supportsOffsetCorrection: false,
   trackAverageRenderTimeForOffsetProjection: false,
   isRN083OrAbove: true,
+  invertedTransformStyle: { transform: [{ scaleY: -1 }] } as const,
+  invertedTransformStyleHorizontal: { transform: [{ scaleX: -1 }] } as const,
```
tmgsca: Just confirming we want this `as const` here

naqvitalha:
Addressed review comments:

- **`as const` on PlatformHelper.web.ts**: Yes, intentional — it's consistent across all four platform helpers (ios, android, web, default). Narrows the transform values to literal types which aligns with RN's `ViewStyle` transform expectations.

- **Early return in `getInvertedTransformStyle`**: Good call — moved the `inverted` guard to the call sites. `getInvertedTransformStyle` now only takes `horizontal` and always returns a style object. (85a62469)

lucasklaassen:
Appreciate you adding this but I notice the test on android checkbox on the PR is left unchecked. Would you like some feedback on how it performs on a real android device?

PR #2146, opened 2026-03-11
fix(scroll): guard against zero window size in bound detection

### LINKED ISSUE #1956: [Android] Infinite list infinitely loading in the background after cold start
## Description
Observed a weird behavior after upgrading to FlashList v2 where the list would continuously load more and more pages under specific circumstances.

We are using a FlashList on our root "Feed" page in a `BottomTabNavigator` and when deep linking to other screens, the `BottomTabNavigator` is mounted so the feed page is always there when you go back.

Problem: When cold starting the app with a deep link to a different screen (outside of the BottomTabNavigator) the FlashList would get mounted and populated in the background, but all it's measurements, both window and items, are coming back with 0 width, 0 height and the computed visible indices are the entire list (`[0, items.length -1]`) which then caused problems.

This happened only on production builds (or after running expo with `--no-dev`)

There might be be mathematical argument that one can fit infinite 0 height items in a 0 height container (`n * 0 <= 0` ✅ ) but from a practical standpoint it may make sense to guard against this?

## Current behavior
1. The `onEndReach` method would be called repeatedly. Since we're using infinite scrolling for this component this resulted in more more and more items being loaded, towards infinity.
1. When going back to the feed the scroll position would be at the bottom of the list

## Expected behavior
1. No `onEndReach` calls are made based off of 0 height calculations
1. Scroll position is maintained at the top of the list when going back

## Reproduction
* Was unable to reproduce outside our app. Including my repro attempt for reference

* Unable to reproduce this issue in dev mode, only prod builds or using `--no-dev`

* Verified that the measurements were coming back as 0 when this occurred via logs:

```ts
    console.log('Window size', listRef?.current?.getWindowSize());
    console.log('Layout 0', listRef?.current?.getLayout(0));
    console.log('Layout 1', listRef?.current?.getLayout(1));
    console.log('Visible indices', listRef?.current?.computeVisibleIndices());
```

Example output
```
I  'Window size', { width: 0, height: 0 }
I  'Layout 0', { x: 0,
     y: 0,
     width: 0,
     height: 0,
     isWidthMeasured: true,
     enforcedWidth: true,
     isHeightMeasured: true }
I  'Layout 1', { x: 0,
     y: 0,
     width: 0,
     height: 0,
     isWidthMeasured: true,
     enforcedWidth: true,
     isHeightMeasured: true }
I  'Visible indicies', { startIndex: 0, endIndex: 91 }
```

https://github.com/siggi-livekindred/flash-list-infinite-loop-repro

## Platform

- [ ] iOS
- [x] Android
- [ ] Web (if applicable)

## Environment

```
System:
  OS: macOS 15.6.1
  CPU: (12) arm64 Apple M2 Pro
  Memory: 134.78 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.18.0
    path: /Users/siggijonsson/.nvm/versions/node/v22.18.0/bin/node
  Yarn: Not Found
  npm:
    version: 10.9.3
    path: /Users/siggijonsson/.nvm/versions/node/v22.18.0/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.16.2
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.0
      - iOS 26.0
      - macOS 26.0
      - tvOS 26.0
      - visionOS 26.0
      - watchOS 26.0
  Android SDK: Not Found
IDEs:
  Android Studio: 2025.1 AI-251.27812.49.2514.14217341
  Xcode:
    version: 26.0.1/17A400
    path: /usr/bin/xcodebuild
Languages:
  Java: Not Found
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli":
    installed: 20.0.2
    wanted: latest
  react:
    installed: 19.1.0
    wanted: 19.1.0
  react-native:
    installed: 0.81.4
    wanted: 0.81.4
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

</details>

**FlashList version:** 2.0.3 (issue also present in 2.1.0)

## Additional context

I know that it's hard to identify the cause without a working repro. Could be a combination of different things in our app (our feed screen is quite complex) that leads to this but hoping for some insights on why FlashList would proceed to operate as normal after it's measured at 0 height.

Workarounds:
* Added `useIsFocused()` safe guards to short circuit `onEndReach` if the screen is not focused.
* Defer rendering of the `FlashList` until it's parent screen component has been measured (via `onLayout`)

## Checklist

naqvitalha:
## Summary
- Adds a `visibleLength <= 0` early return in `useBoundDetection` to skip `onEndReached`/`onStartReached` when the list has no measurable size
- Prevents infinite loading when the list is mounted off-screen (e.g., background tab) where all measurements are 0

Fixes #1956

## Test plan
- [x] Verify `onEndReached` no longer fires when list is mounted in a background/hidden tab
- [x] Verify `onEndReached` still fires correctly during normal scrolling
- [x] Unit tests pass (`yarn test`)

### DIFF
--- a/src/recyclerview/hooks/useBoundDetection.ts
+++ b/src/recyclerview/hooks/useBoundDetection.ts
@@ -84,6 +84,14 @@ export function useBoundDetection<T>(
         (isHorizontal ? contentSize.width : contentSize.height) +
         recyclerViewManager.firstItemOffset;
 
+      // Skip bound detection if the window has no measurable size.
+      // This can happen when the list is mounted off-screen (e.g., in a
+      // background tab) and all measurements come back as 0, which would
+      // incorrectly trigger onEndReached/onStartReached.
+      if (visibleLength <= 0) {
+        return;
+      }
+
       // Check if we're near the end of the list
       if (onEndReached) {
         const onEndReachedThreshold = onEndReachedThresholdProp ?? 0.5;


naqvitalha:
@agent review this PR

gantonious (approved):

Issue #2175, opened 2026-03-20
Dynamic item height causes items to collapse/overlap during fast scroll


Jianlong-Nie:
## Description

When using FlashList with dynamic item heights, quickly scrolling up/down causes content to jump, collapse, and overlap each other.

## Steps to Reproduce

1. Render a FlashList with dynamic item heights (items have varying/unpredictable heights)
2. Scroll quickly up and down

## Expected Behavior

Items maintain their correct heights and positions while scrolling.

## Actual Behavior

Items jump, collapse, and overlap each other during fast scroll due to incorrect height estimations for dynamic items.

## Root Cause (suspected)

FlashList relies on `estimatedItemSize` for layout calculations. When item heights are dynamic and differ significantly from the estimated size, fast scrolling causes FlashList to miscalculate item positions, resulting in visible layout shifts and overlapping content.

## Platform

- iOS
- Android

## FlashList Version

2.3.0

naqvitalha:
Can you provide a sample where this can be reproduced? That will help us look into it.

Issue #1914, opened 2025-09-18
Unable to find viewState for tag 10254. Surface stopped: false


renyufei8023:
## Description

When FlashList is pulled up to load more or slid down, the app will crash

## Expected behavior

I hope it won't collapse while sliding

## Platform

- [x] Android

## Environment

<details><summary>React Native info output:</summary>

```
System:
  OS: macOS 15.6.1
  CPU: (4) x64 Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
  Memory: 53.39 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.16.0
    path: /usr/local/opt/node@22/bin/node
  Yarn:
    version: 1.22.22
    path: /usr/local/opt/node@22/bin/yarn
  npm:
    version: 10.9.2
    path: /usr/local/opt/node@22/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.16.2
    path: /usr/local/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.5
      - iOS 18.5
      - macOS 15.5
      - tvOS 18.5
      - visionOS 2.5
      - watchOS 11.5
  Android SDK: Not Found
IDEs:
  Android Studio: 2024.3 AI-243.25659.59.2432.13423653
  Xcode:
    version: 16.4/16F6
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.11
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli":
    installed: 20.0.0
    wanted: 20.0.0
  react:
    installed: 19.1.0
    wanted: 19.1.0
  react-native:
    installed: 0.81.4
    wanted: 0.81.4
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: false

```

</details>

**FlashList version:2.0.3

## Additional context

I think it may be related to removeClippedSubviews

## Checklist

- [x] I've searched existing issues and couldn't find a duplicate
- [ ] I've provided a minimal reproduction (Expo Snack preferred)
- [x] I'm using the latest version of @shopify/flash-list
- [x] I've included all required information above

[...]

PR #2199, opened 2026-03-23
chore(ci): pass explicit issue number to agent-fix prompt

naqvitalha:
## Description

The agent-fix workflow prompt said "Fix this GitHub issue" without specifying which one, relying on claude-code-action to inject context automatically. In run #23455139708, the agent was assigned issue #1914 but drifted over 136 turns to fix issue #2175 instead — a completely different bug.

Now the issue number and title are passed explicitly via `${{ github.event.issue.number }}` and `${{ github.event.issue.title }}`, matching how agent-triage already works.

Simple template variable addition. Compare with agent-triage.yml which already does this correctly.

## Test plan
- [ ] Next agent-fix run targets the correct issue


naqvitalha:
Superseded by #2200 (combined PR)

Issue #1816, opened 2025-08-02
Scroll position changes when navigating back to a list in react navigation stack


moh3n9595:
## Current behavior

Upon navigating backwards, the list's scroll position is different from where it was originally.

## Expected behavior

The scroll position should not change.

## Platform:

- [x] Web
- [ ] iOS
- [ ] Android

## Environment

2.0.1

ericpoulinnz:
@naqvitalha sorry for the tag but this is a longstanding issue (#635 ) which V2 has not fixed. It makes for a terrible UX for a very common use case involving lists.

SergioAN007:
Just set navigator option:  screenOptions={{detachPreviousScreen: false}}

spyshower:
Same issue here, but it happens when pushing a new screen.

vatsal-bk:
Any updates on this? I am facing same issue

mayconmesquita:
Thanks! This solved the issue for me. More details in the [docs](https://reactnavigation.org/docs/stack-navigator/#detachinactivescreens).

> Just set navigator option: screenOptions={{detachPreviousScreen: false}}

takumatakuma:
We hit this in a production Expo app (expo-router `Stack` → `@react-navigation/native-stack` on web) and traced it to a specific missing guard. Sharing the measurements and a patch in case it helps.

## It is FlashList-specific, not a browser or navigator behaviour

Four scrollers rendered inside the **same** subtree, toggled `display: none` → `flex` once (Chromium via Playwright, `@shopify/flash-list@2.0.2`, `react-native-web@0.21.2`):

| scroller | before hide | after show |
|---|---|---|
| plain DOM `overflow:auto` div | 2000 | **2000** |
| react-native-web `ScrollView` | 2000 | **2000** |
| react-native-web `FlatList` | 2000 | **2000** |
| `@shopify/flash-list` | 2000 | **0** |

The browser restores `scrollTop` across a `display:none` round trip by itself. The position is lost only because the list shrinks its own content to nothing while hidden.

## Mechanism

1. `NativeStackView.js:94` keeps the unfocused screen mounted with `display: none`.
2. ResizeObserver fires; `react-native-web`'s `useElementLayout` re-measures the hidden node and dispatches `onLayout({width: 0, height: 0})`.
[...]

PR #2269, opened 2026-05-04
fix(layout): skip layout updates when container size is 0x0

### LINKED ISSUE #2231: Rendering of all data when navigating with @react-navigation/stack
## Description

Using @react-navigation/stack, pushing on top of a screen with FlashList makes FlashList render all list items.

## Current behavior

FlashList re-renders all of its ListItems, even those that aren't in the viewport.

https://github.com/user-attachments/assets/ce1c658d-a1a7-48e9-ba09-c276cec17d14

## Expected behavior

No additional renders should happen.

## Reproduction

https://snack.expo.dev/@sharabai/flashlist-renderitem-bug-with-react-stack-navigation?platform=web

## Platform

- [ ] iOS
- [ ] Android
- [x] Web (if applicable)

## Environment

**FlashList version:** 2.3.0

## Additional context

The issue is tightly linked to #1816. In the thread under that issue you can find a workaround that involves manually setting some prop in Navigator.

The core problem is that React Navigation on web makes FlashList's dimensions 0x0 when the screen is being pushed on, which makes FlashList recalculate the height of its children and because of it all sorts of problems arise.

Proposed fix that was used at https://github.com/Expensify/App as a patch is doing an early return when FlashList's size is 0x0. 
It makes it a lot easier to use in real-life production app without worrying about passing some props to navigation, which can influence perf negatively and you have to remember to do it each time. Working on such a huge projects, it's a lot easier to solve it at the level of FlashList itself. 
```diff
diff --git a/src/recyclerview/RecyclerView.tsx b/src/recyclerview/RecyclerView.tsx
index 34722d48..c14346f1 100644
--- a/src/recyclerview/RecyclerView.tsx
+++ b/src/recyclerview/RecyclerView.tsx
@@ -166,6 +166,10 @@ const RecyclerViewComponent = <T,>(
     if (internalViewRef.current && firstChildViewRef.current) {
       // Measure the outer container size and inner container layout
       const outerViewSize = measureParentSize(internalViewRef.current);
+      if (outerViewSize.width === 0 && outerViewSize.height === 0) {
+        containerViewSizeRef.current = outerViewSize;
+        return;
+      }
       const firstChildViewLayout = measureFirstChildLayout(
         firstChildViewRef.current,
         internalViewRef.current
@@ -204,6 +208,12 @@ const RecyclerViewComponent = <T,>(
     if (pendingChildIds.size > 0) {
       return;
     }
+    if (
+      containerViewSizeRef.current?.width === 0 &&
+      containerViewSizeRef.current?.height === 0
+    ) {
+      return;
+    }
     const layoutInfo = Array.from(refHolder, ([index, viewHolderRef]) => {
       const layout = measureItemLayout(
         viewHolderRef.current!,
```

## Checklist

sharabai:
## Description

Fixes #2231

When FlashList is mounted inside a screen rendered by `@react-navigation/stack` on web, pushing a new screen on top of it transiently measures FlashList's container as `0x0`. The two layout `useLayoutEffect`s in `RecyclerView` then run with those bogus dimensions: the first overwrites the window/layout params with zeros, and the second walks every `viewHolderRef` and rewrites item layouts. The downstream effect is that every `renderItem` re-runs — including offscreen items — which defeats recycling and causes the perf regression in the linked Snack repro.

This PR adds two early returns inside `src/recyclerview/RecyclerView.tsx`:

1. In the post-mount layout effect: if the outer container measures `0x0`, store that size in `containerViewSizeRef` and bail out before calling `recyclerViewManager.updateLayoutParams(...)`.
2. In the per-item layout effect: if `containerViewSizeRef.current` is `0x0`, bail out before iterating `refHolder` and rewriting layouts.

Once the screen is brought back into view and the container regains a non-zero size, the normal layout path resumes on the next layout pass — no items are dropped or re-mounted.

This is the same patch Expensify currently ships against `@shopify/flash-list@2.3.0` for the same root cause discussed in #1816, and it removes the need for consumers to remember to toggle a Navigator-level workaround at every call site.

- [ ] Confirm the first early return still writes `containerViewSizeRef.current = outerViewSize` before bailing, so the second effect can detect the `0x0` state.
- [ ] Reproduce #2231 against `main` using the Snack from the issue, then re-run against this branch and verify offscreen items stop re-rendering when a stack screen is pushed on top.
- [ ] Verify there's no regression on lists that legitimately mount with non-zero size on the first measurement (mobile + web), i.e. the fast path is unchanged.
- [ ] Sanity-check that layout converges on the next non-zero measurement after the container is restored — items shouldn't be left half-measured.

## Screenshots or videos 

After-fix recording: 

https://github.com/user-attachments/assets/124b5032-c3ea-423c-8fbc-037097747295

### DIFF
--- a/src/recyclerview/RecyclerView.tsx
+++ b/src/recyclerview/RecyclerView.tsx
@@ -166,6 +166,10 @@ const RecyclerViewComponent = <T,>(
     if (internalViewRef.current && firstChildViewRef.current) {
       // Measure the outer container size and inner container layout
       const outerViewSize = measureParentSize(internalViewRef.current);
+      if (outerViewSize.width === 0 && outerViewSize.height === 0) {
+        containerViewSizeRef.current = outerViewSize;
+        return;
+      }
       const firstChildViewLayout = measureFirstChildLayout(
         firstChildViewRef.current,
         internalViewRef.current
@@ -202,6 +206,12 @@ const RecyclerViewComponent = <T,>(
   // eslint-disable-next-line react-hooks/exhaustive-deps
   useLayoutEffect(() => {
     if (pendingChildIds.size > 0) {
+      return;
+    }
+    if (
+      containerViewSizeRef.current?.width === 0 &&
+      containerViewSizeRef.current?.height === 0
+    ) {
       return;
     }
     const layoutInfo = Array.from(refHolder, ([index, viewHolderRef]) => {
--- a/src/recyclerview/hooks/useRecyclerViewController.tsx
+++ b/src/recyclerview/hooks/useRecyclerViewController.tsx
@@ -238,8 +238,14 @@ export function useRecyclerViewController<T>(
       }: ScrollToOffsetParams) => {
         const { horizontal } = recyclerViewManager.props;
         if (scrollViewRef.current) {
-          // Adjust offset for RTL layouts in horizontal mode
-          if (I18nManager.isRTL && horizontal) {
+          // Adjust offset for RTL layouts in horizontal mode.
+          // Gated on hasLayout(): the RTL math reads child container and
+          // window size, which throw while the list is hidden (0x0).
+          if (
+            I18nManager.isRTL &&
+            horizontal &&
+            recyclerViewManager.hasLayout()
+          ) {
             // eslint-disable-next-line no-param-reassign
             offset =
               adjustOffsetForRTL(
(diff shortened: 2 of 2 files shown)


sharabai:
CLA is signed.

sharabai:

sharabai:
@naqvitalha I've added a new commit, which guards against crashes caused by the early returns. 
At Expensify, we've had some issues because of the early returns, so @VickyStash did some debugging and fixed the places that were causing this. This PR is now tested on a pretty big project. That is to say the PR is battle tested. I've added her contribution to this PR too.

sharabai:
The CI is stalled for some reason. I'm not sure what needs to be done. 
@amit-hinge could you take a look? Not sure who to tag here.

amit-hinge:
@naqvitalha is the best person here to help you with!

sharabai:
@naqvitalha friendly bump :)

Issue #2070, opened 2026-01-23
List flickering if `maintainVisibleContentPosition.startRenderingFromBottom` is set to true


VickyStash:
## Description

When `maintainVisibleContentPosition.startRenderingFromBottom` is set to true, only the bottom item is rendered at first. This is causing a flickering effect during list rendering.
It seems to be okay if `maintainVisibleContentPosition.startRenderingFromBottom` is set to false.

## Current behavior

When `maintainVisibleContentPosition.startRenderingFromBottom` is set to true, only the bottom item is rendered at first.

## Expected behavior

When `maintainVisibleContentPosition.startRenderingFromBottom` is set to true, items to fill the displayed area are rendered.

## Reproduction

**Expo Snack or minimal reproduction link:**

The issue is reproduced in the `fixture` app
https://github.com/Shopify/flash-list/tree/921847d35cca043f438a441e9348bad358ffc0fd/fixture

(videos are slowed down so it's easier to see the issue)

https://github.com/user-attachments/assets/db5fdea6-ce47-4b0c-bea6-4b8bae075f84

https://github.com/user-attachments/assets/3e4a4b0d-de06-417f-9af0-0d996996dbe4

## Platform

- [x] iOS
- [x] Android
- [x] Web (if applicable)

## Environment

<details><summary>React Native info output:</summary>

```
System:
  OS: macOS 26.2
  CPU: (10) arm64 Apple M2 Pro
  Memory: 125.64 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.19.5
    path: ~/.nvm/versions/node/v20.19.5/bin/node
  Yarn:
    version: 1.22.19
    path: /opt/homebrew/bin/yarn
  npm:
    version: 10.8.2
    path: ~/.nvm/versions/node/v20.19.5/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods: Not Found
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2025.2 AI-252.28238.7.2523.14688667
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.9
    path: /usr/bin/javac
[...]

PR #2290, opened 2026-05-21
fix: scroll to index autoscroll race

isekovanic:
## Description

This PR fixes a race between imperative `scrollToIndex` and the MVCP autoscroll to bottom mechanism in v2. When `autoscrollToBottomThreshold` is set and the user is near the bottom of the list, calling [`scrollToIndex`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L320) imperatively can be hijacked by the autoscroll to bottom path mid scroll animation. The list briefly lands at the target index, then snaps back to the bottom.

This typically reproduces when list items are not fully stable, i.e they need to be remeasured/corrected often. In other words, something that would actually falsely trigger autoscrolling.

I've found this to be reliably reproducible in any chat style list with an actual `autoscrollToBottomThreshold` set to its MVCP that also has unstable items (i.e images or GIFs, but also items which might require additional correction internally so basically anything that would trigger any type of change in `contentSize`):

1. Open the list (lands at bottom)
2. Scroll up far enough that the target index has rendered and been measured at least once
3. Scroll back down to the bottom and sit still for >=100ms
4. Trigger `listRef.current?.scrollToIndex({ index: targetIndex, animated: true, viewPosition: 0.5 })` - typically from an effect or `setTimeout(0)` callback that's downstream of an unrelated state update (e.g., tapping a quoted message that first dispatches a "targeted message" state change)
5. List animates to `targetIndex`, then snaps back to the bottom

The race fires reliably only on the first attempt after step 3. Subsequent calls behave correctly: the race fired `scrollToEnd` refreshes the tracked anchor (via `computeFirstVisibleIndexForOffsetCorrection`), so subsequent `applyOffsetCorrection` diff calculations evaluate to ~0 and don't emit the stray `scrollBy`. A long enough wait at the bottom can make the snapshot stale again however, via background remeasurement and reproduce on a later attempt.

After a ton of debugging, I've found the root cause of this to actually be the two MVCP related subsystems interacting without coordinating:

**1. Autoscroll-to-bottom** lives in [`useBoundDetection.ts`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts). It maintains a sticky ref [`pendingAutoscrollToBottom`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L28), set `true` by [`checkBounds`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L55) whenever the user is within `autoscrollToBottomThreshold` viewports of the bottom. [`runAutoScrollToBottomCheck`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L143) fires `scrollToEnd` via [`requestAnimationFrame`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L154), driven by two effects:
- The [data change effect](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L173), keyed on `[data, runAutoScrollToBottomCheck, windowHeight, windowWidth]`.
- The [content geometry effect](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L178), keyed on `[contentHeight, contentWidth, recyclerViewManager.firstItemOffset, …]`, gated by a [100ms `lastCheckBoundsTime` guard](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L179) that I'm guessing is there to avoid firing during finger scroll.

**2. Anchor MVCP** lives in [`applyOffsetCorrection`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L119), runs on every layout commit and applies a compensating [`scrollBy(diff)`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L186) when its [tracked anchor item](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L58)'s cached layout snapshot diverges from the live layout. Gated by [`pauseOffsetCorrection.current`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L53). The anchor snapshot is refreshed at the end of each `applyOffsetCorrection` run via [`computeFirstVisibleIndexForOffsetCorrection`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L92) and at scroll momentumend via the same function.

So, the timeline of events looks something like this from my analysis:

- User is at the bottom; `pendingAutoscrollToBottom.current === true`. The user has been sitting still long enough that background remeasurement of items (async images, font metric resolution, estimates resolving into real measurements as items leave/enter the render window) has drifted some items' cached layouts relative to the snapshot captured by `computeFirstVisibleIndexForOffsetCorrection` at the last momentum end.
- An unrelated rerender commits typically from a state update in the integrating app (e.g., highlighting the targeted message). `pauseOffsetCorrection.current` is still `false` because `scrollToIndex` hasn't run yet - it lives downstream in a `setTimeout(0)` / effect / event-handler tick.
- During that commit, `applyOffsetCorrection` evaluates with the stale anchor snapshot, computes a nonzero `diff` and fires [`scrollAnchorRef.current?.scrollBy(diff)`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L186)
- The stray scroll mutates `contentHeight` / `firstItemOffset` over the next render, triggering the content change `useEffect` in `useBoundDetection`. `pendingAutoscrollToBottom.current` is still `true` (the >100ms `lastCheckBoundsTime` guard passes), then  `runAutoScrollToBottomCheck` fires, then `scrollToEnd` queued via `requestAnimationFrame`
- A tick later, `scrollToIndex` finally runs and sets [`setOffsetProjectionEnabled(false)`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L335) + `pauseOffsetCorrection.current = true`. Two scroll animations are now in flight; `scrollToEnd` typically wins.

### Fix

A single early return at the top of [`runAutoScrollToBottomCheck`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useBoundDetection.ts#L143-L156): bail when a programmatic `scrollToIndex` is in flight, signalled by FlashList's own [`recyclerViewManager.isOffsetProjectionEnabled`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/RecyclerViewManager.ts#L53-L55) getter, backed by [`engagedIndicesTracker.enableOffsetProjection`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/helpers/EngagedIndicesTracker.ts#L67) (default `true`).

```ts
const runAutoScrollToBottomCheck = useCallback(() => {
+ if (!recyclerViewManager.isOffsetProjectionEnabled) {
+   return;
+ }
  if (pendingAutoscrollToBottom.current) {
    pendingAutoscrollToBottom.current = false;
    requestAnimationFrame(() => {
      const shouldAnimate =
        recyclerViewManager.props.maintainVisibleContentPosition
          ?.animateAutoScrollToBottom ?? true;
      scrollViewRef.current?.scrollToEnd({
        animated: shouldAnimate && !recyclerViewManager.ignoreScrollEvents,
      });
    });
  }
}, [requestAnimationFrame, scrollViewRef, recyclerViewManager]);
```

`isOffsetProjectionEnabled` is already toggled `false` at [the start of `scrollToIndex`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L335) and back to `true` ~200–300ms after the scroll settles ([here](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/hooks/useRecyclerViewController.tsx#L487-L488)). That window covers the entire `scrollToIndex` operation and its settling tail, during which `applyOffsetCorrection` can continue to emit corrective scrolls. If it fits the usecase better I can possibly introduce a separate ref that is going to have purely this as its responsibility.

The autoscroll to bottom contract should be *"if the user is at the bottom and new content arrives, stay at the bottom."* When an imperative scroll has explicitly reassigned the scroll position, "stay at the bottom" no longer applies for the duration of that operation. Suppressing autoscroll while imperative scrolling owns the position should be the correct semantic in my opinion.

I've also included a new fixture that reproduces this consistently (just load all of the items up until the middle and try scrolling to bottom and then to the item using the buttons. Needed to do this because of how stable the example is and it wouldn't reproduce on items that don't particularly drift at any point. 

- [ ] The guard predicate (`!recyclerViewManager.isOffsetProjectionEnabled`) is the right signal vs. alternatives like `pauseOffsetCorrection.current` or a new dedicated flag
- [ ] No other call site of `runAutoScrollToBottomCheck` or path into autoscroll is unintentionally suppressed by this guard. `isOffsetProjectionEnabled` is only flipped inside `scrollToIndex`, so the window is precisely the imperative scroll lifetime + settle
- [ ] Confirm the "stay at bottom on new content arriving during a programmatic scroll" behavior change is acceptable - during the 200-300ms settling tail of a `scrollToIndex`, newly arrived data won't trigger autoscroll. Autoscroll resumes when the flag flips back
- [ ] Confirm the issue is reproducible through the fixture screen

Before:

https://github.com/user-attachments/assets/6e192187-60de-40fd-99c0-efc38e76bc1a

(I forgot to add feedback on clicking the buttons but essentially I'm just pressing on the `scrollToIndex` button and it succeeds only once)

After:

https://github.com/user-attachments/assets/e9054b92-c518-431e-b0a2-32a064f29570

(same drill, except it succeeds all times)

### DIFF
--- a/fixture/react-native/src/ChatUnstableItems.tsx
+++ b/fixture/react-native/src/ChatUnstableItems.tsx
@@ -1,0 +1,412 @@
+import React, {
+  useState,
+  useCallback,
+  useMemo,
+  useRef,
+  useEffect,
+} from "react";
+import {
+  View,
+  Text,
+  Image,
+  StyleSheet,
+  Pressable,
+  SafeAreaView,
+  StatusBar,
+} from "react-native";
+import {
+  FlashList,
+  FlashListRef,
+  useRecyclingState,
+} from "@shopify/flash-list";
+
+interface ChatMessage {
+  id: string;
+  text: string;
+  sender: "user" | "other";
+  timestamp: Date;
+}
+
+const messageTexts = [
+  "Hey, how are you?",
+  "Hey, how are you? Hey, how are you? Hey, how are you? Hey, how are you?",
+  "What's up?",
+  "What's up? What's up? What's up?",
+  "Did you see that new movie?",
+  "I'm working on a new project",
+  "Let's catch up soon!",
+  "Have you tried the new restaurant downtown?",
+  "Have you tried the new restaurant downtown?, Have you tried the new restaurant downtown?, Have you tried the new restaurant downtown?, Have you tried the new restaurant downtown?",
+  "Just finished reading an amazing book",
+  "Can you send me that file?",
+  "Are we still meeting tomorrow?",
+  [
+    "Are we still meeting tomorrow?",
+    "Are we still meeting tomorrow?",
+    "Are we still meeting tomorrow?",
+    "Are we still meeting tomorrow?",
+    "Are we still meeting tomorrow?",
+    "Are we still meeting tomorrow?",
+    "Are we still meeting tomorrow?",
+  ].join(" "),
+  "The weather is beautiful today!",
+  "The weather is beautiful today! The weather is beautiful today!",
+];
+
+export function ChatUnstableItems() {
+  const [messages, setMessages] = useState<ChatMessage[]>(() =>
+    generateInitialMessages(300)
+  );
+  const listRef = useRef<FlashListRef<ChatMessage>>(null);
+
+  const addMessageAtTop = useCallback(() => {
+    const newMessages = Array.from({ length: 50 }, (_, i) =>
+      generateRandomMessage(i)
+    );
+    setMessages((prev) => [...newMessages, ...prev]);
+  }, []);
+
+  // const addMessageAtBottom = useCallback(() => {
+  //   const newMessage = generateRandomMessage(messages.length);
+  //   setMessages((prev) => [...prev, newMessage]);
+  //   // eslint-disable-next-line react-hooks/exhaustive-deps
+  // }, []);
+
+  // Mirror stream-chat-react-native's "tap a reply" timing pattern:
+  // the tap schedules a state update; the actual scrollToIndex runs on a
+  // later tick via setTimeout(0). This separates the re-render (where
+  // applyOffsetCorrection can fire its stray scrollBy with the anchor's
+  // stale layout snapshot) from the scrollToIndex call (which would
+  // otherwise have already set pauseOffsetCorrection.current = true and
+  // suppressed the stray scrollBy).
+  const [scrollTarget, setScrollTarget] = useState<number | null>(null);
+
+  useEffect(() => {
+    if (scrollTarget === null) return undefined;
+    const id = setTimeout(() => {
+      listRef.current?.scrollToIndex({
+        index: scrollTarget,
+        animated: true,
+        viewPosition: 0.5,
+      });
+      setScrollTarget(null);
+    }, 0);
+    return () => clearTimeout(id);
+  }, [scrollTarget]);
+
+  const scrollToItem25 = useCallback(() => {
+    setScrollTarget(150);
+  }, []);
+
+  const scrollToBottom = useCallback(() => {
+    listRef.current?.scrollToEnd({ animated: true });
+  }, []);
+
+  const renderItem = useCallback((info: { item: ChatMessage }) => {
+    return <MessageBubble message={info.item} />;
+  }, []);
+
+  const maintainVisibleContentPositionConfig = useMemo(
+    () => ({
+      animateAutoScrollToBottom: true,
+      autoscrollToBottomThreshold: 1,
+      startRenderingFromBottom: true,
+    }),
+    []
+  );
+
+  const listHeaderComponent = useMemo(
+    () => (
+      <View style={styles.header}>
+        <Text style={styles.headerTitle}>Chat Example</Text>
+      </View>
+    ),
+    []
+  );
+
+  const keyExtractor = useCallback((item: ChatMessage) => item.id, []);
+
+  return (
+    <SafeAreaView style={styles.safeArea} testID="ChatUnstableItemsScreen">
+      <StatusBar barStyle="dark-content" />
+      <View style={styles.container}>
+        <View style={styles.header}>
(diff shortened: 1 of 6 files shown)


isekovanic:
Hey @naqvitalha , just signed the CLA but I'm not sure how to rerun the job as github won't give me an option to do so

isekovanic:
Ah nevermind, I was simply supposed to comment. My bad !

amit-hinge (approved):

naqvitalha (approved):

naqvitalha:
Thanks for fixing this. Looks good.

isekovanic:
Thanks a bunch for the express review @naqvitalha !

Also, as some hindsight here - I've noticed that [this issue](https://github.com/Shopify/flash-list/issues/2070) as well as the one where on Android on first render the list would visibly scroll to bottom (cannot find it anymore) have also disappeared, at least within our SDK ! Now, I have no idea why that might be (seemingly unrelated problems) but perhaps it makes sense for the reporters to retest these in case it was a lucky fix there as well.

PR #2318, opened 2026-06-11
fix(layout): keep layout table monotonic on initial scroll

### LINKED ISSUE #2307: Flashlist v2 initialScrollIndex renders wrong items when item size exceeds 200
## Description

`FlashList` v2 displays wrong items when `initialScrollIndex` is set and the actual item size exceeds 200px (FlashList's internal `averageWindow` default). The list opens showing items roughly 83 indices past the requested index.

## Current behavior

With `initialScrollIndex={250}` and items 300px tall, the list opens showing Item 333 instead of Item 250.

The root cause is a race condition in FlashList's layout table construction:

1. `applyInitialScrollAdjustment` runs before any items are measured, building item positions using the default `averageWindow` of 200px. Item 250 is placed at `250 × 200 = 50,000px` (should be `250 × 300 = 75,000px`).
2. Item 250 renders at the wrong position, measures at 300px, and `modifyLayout` runs `recomputeLayouts(250→500)` — but this anchors items 251+ to item 249's stale 200px-based position (`249 × 200 = 49,800px`). Items 251+ are now at ~50,400px+.
3. `applyInitialScrollAdjustment` fires again with the updated `averageWindow=300` and correctly places items 0–250 at `i × 300px`. Item 250 moves to 75,000px. **However, items 251+ are not touched** — they retain their wrong positions from step 2.
4. The layout table now has a backward jump at the 250→251 boundary (`75,000 → 50,400`), making it non-monotonic.
5. FlashList binary-searches the broken table to find which items to render at scroll position 75,000px. The search starts at mid=249, sees 74,700 < 75,000, goes **right** — and never checks item 250 (which is to the left). It finds item 333 (at position ~75,000 in the broken layout) and renders that instead.

The bug only manifests when actual item size **exceeds 200px**. With smaller items (< 200px), the corrective pass moves items to a *lower* pixel position than what `modifyLayout` set, creating a forward gap that keeps the table monotonically sorted.

## Expected behavior

The list should open with Item 250 visible at the top.

## Reproduction

**Expo Snack:** https://snack.expo.dev/TgCTAo_9hg8zsXmGYrlya

```jsx
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';
import { FlashList } from '@shopify/flash-list';

const ITEM_SIZE = 300;
const INITIAL_INDEX = 250;
const DATA = Array.from({ length: 600 }, (_, i) => i);

  return (
    <View style={{ flex: 1 }}>
      <View style={styles.header}>
        <Text style={styles.title}>Expected: Item {INITIAL_INDEX} at top</Text>
      </View>
      <FlashList
        data={DATA}
        renderItem={({ item }) => (
          <View style={styles.item}>
            <Text style={styles.label}>Item {item}</Text>
          </View>
        )}
        keyExtractor={(item) => String(item)}
        initialScrollIndex={INITIAL_INDEX}
      />
    </View>
  );
}

  header: { padding: 16, backgroundColor: '#f5f5f5', borderBottomWidth: 1, borderBottomColor: '#ddd' },
  title: { fontSize: 16, fontWeight: 'bold' },
  item: { height: ITEM_SIZE, justifyContent: 'center', alignItems: 'center', borderBottomWidth: 1, borderBottomColor: '#eee' },
  label: { fontSize: 32, fontWeight: 'bold' },
});
```

Platform
 iOS
 Android
Environment

React Native: 0.81.6
FlashList version: 2.3.1
</details>

Additional context

This bug was introduced in FlashList v2 when estimatedItemSize was removed from the public API — a deliberate design decision to eliminate a common source of misconfiguration.

Root of the problem: applyInitialScrollAdjustment runs a self-correcting two-pass loop: the first pass estimates positions using the default averageWindow (200px), and the second corrects them after items are measured. However, modifyLayout fires between the two passes and anchors items after initialScrollIndex to the wrong predecessor position. The second corrective pass then fixes items 0–initialScrollIndex but stops there — leaving a non-monotonic boundary at initialScrollIndex + 1 that breaks the binary search used to select visible items.

Workaround: restore estimatedItemSize and use it to seed the averageWindow so the very first pass produces correct positions, preventing the race condition from occurring. This is not ideal — it reintroduces a prop the maintainers intentionally removed — and there is likely a better internal fix (for example, after the corrective recomputeLayouts(0 → initialScrollIndex), propagating correct positions forward into the items that modifyLayout had incorrectly anchored). I haven't worked out the cleanest approach on the library side, but wanted to document the root cause in detail so it can inform the right fix.

Checklist
 I've searched existing issues and couldn't find a duplicate
 I've provided a minimal reproduction (Expo Snack preferred)
 I'm using the latest version of @shopify/flash-list
 I've included all required information above

ahmdshrif:
## Description

Fixes #2307.

With `initialScrollIndex` set and items whose real size exceeds the layout managers' default 200px average estimate, the list opens showing the wrong items (e.g. item 333 instead of 250 with 300px items).

Root cause: `applyInitialScrollAdjustment` recomputes layouts only over `0..initialScrollIndex` before reading the target offset. When the average estimate has been updated by a first measurement, that pass moves items `0..initialScrollIndex` to new (larger) positions, while items beyond `initialScrollIndex` keep positions derived from the old estimate. The layout table becomes non-monotonic at the `initialScrollIndex` boundary (e.g. `75,000 → 50,400`), and the binary search over item positions then resolves the scroll offset to the wrong region of the table.

The fix recomputes the entire table in this pass, keeping it monotonic. This path only runs during the initial progressive-render passes (`renderProgressively` until the first layout completes), and the recompute is a simple position chain, so the cost class is unchanged from the existing `0..initialScrollIndex` pass.

## Reviewers’ hat-rack

- The measurement path (`modifyLayout`) already recomputes to the end of the table when items are appended; only this initial-scroll corrective pass stopped early.
- Items smaller than the estimate don't trigger the bug (the corrective pass moves items to *lower* positions, which keeps the table sorted) — matching the issue's observation that only sizes > 200px reproduce it.

## Test plan

- New test `src/__tests__/InitialScrollIndexLargeItems.test.tsx`: 500 items measuring 300px with `initialScrollIndex={250}`.
  - **Before the fix** it fails exactly as reported in #2307: items 333–335 are rendered instead of 250.
  - **After the fix** item 250 renders.
- `yarn test --forceExit` — 15 suites, 188 tests pass.
- `yarn type-check` and `yarn lint` — clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Fix wrong items rendered with `initialScrollIndex` when item size exceeds the default average estimate
+  - https://github.com/Shopify/flash-list/issues/2307
 - Introduce optional view offset param for initial scroll position
   - https://github.com/Shopify/flash-list/pull/1870
 - Add sticky header offset and sticky header backgrounds
--- a/src/__tests__/InitialScrollIndexLargeItems.test.tsx
+++ b/src/__tests__/InitialScrollIndexLargeItems.test.tsx
@@ -1,0 +1,63 @@
+import React from "react";
+import { Text } from "react-native";
+import "@quilted/react-testing/matchers";
+import { render } from "@quilted/react-testing";
+
+import { FlashList } from "..";
+
+// Items measure at 300px — larger than the layout managers' default
+// 200px average estimate. Reproduces #2307: the initial-scroll corrective
+// pass used to leave items beyond initialScrollIndex at stale positions,
+// breaking the layout table's monotonicity and rendering the wrong items.
+jest.mock("../recyclerview/utils/measureLayout", () => {
+  const originalModule = jest.requireActual(
+    "../recyclerview/utils/measureLayout"
+  );
+  return {
+    ...originalModule,
+    measureParentSize: jest.fn().mockImplementation(() => ({
+      width: 399,
+      height: 899,
+    })),
+    measureFirstChildLayout: jest.fn().mockImplementation(() => ({
+      x: 0,
+      y: 0,
+      width: 399,
+      height: 899,
+    })),
+    measureItemLayout: jest.fn().mockImplementation(() => ({
+      x: 0,
+      y: 0,
+      width: 399,
+      height: 300,
+    })),
+  };
+});
+
+describe("initialScrollIndex with items larger than the default estimate", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.useFakeTimers();
+  });
+
+  it("renders the requested item when item size exceeds the average estimate", () => {
+    const data = Array.from({ length: 500 }, (_, index) => index);
+
+    const result = render(
+      <FlashList
+        data={data}
+        initialScrollIndex={250}
+        overrideProps={{ initialDrawBatchSize: 1 }}
+        drawDistance={0}
+        renderItem={({ item }) => <Text>{item}</Text>}
+      />
+    );
+
+    jest.runAllTimers();
+
+    // The list should open at item 250 — not ~83 indices further down,
+    // which is what a non-monotonic layout table produces.
+    expect(result).toContainReactComponent(Text, { children: 250 });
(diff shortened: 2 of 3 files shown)


ahmdshrif:
I've signed the CLA.

PR #2331, opened 2026-06-17
fix(scroll): correct inverted list rendering and wheel direction on web

HuuNguyen312:
## Description

**Fixes inverted list rendering and mouse-wheel direction on web** (e.g. Electron). Native is unaffected.

On web, inverted lists are flipped with `transform: scaleY(-1)` (`scaleX(-1)` when horizontal). This broke two things:

### 1. Rendering froze (~20 items, then blank)

`measureFirstChildLayout` read mirrored `getBoundingClientRect` coordinates, so `firstItemOffset` grew with scroll instead of staying `~0`. The internal `scrollOffset` (`y - firstItemOffset`) went negative → `getVisibleLayouts` clamped to index `0` → the engaged index froze → `renderItem` stopped being called while scrolling.

**Fix:** detect the `scaleY(-1)` transform and measure from the **bottom edge** when flipped, so `firstItemOffset` stays `~0` (matching non-inverted).

### 2. Mouse wheel scrolled the wrong way

The native wheel delta is not flipped by the container transform, so it scrolled opposite to the content.

**Fix:** a web-only `useInvertedWheelFix` hook re-inverts the wheel delta on the scroll node (`-= deltaY`, or `-= deltaX` when horizontal). Setting `scrollTop` / `scrollLeft` still fires the scroll event, so virtualization keeps working.

> **Native behavior is unchanged** — `measureLayout.ts` and the native hook variant (`useInvertedWheelFix.ts`) are no-ops / untouched.

**Affected package:** `@shopify/flash-list`

**Changed files**

| File | Change |
|---|---|
| `src/recyclerview/utils/measureLayout.web.ts` | add `isVerticallyFlipped`, measure from bottom when flipped |
| `src/recyclerview/hooks/useInvertedWheelFix.web.ts` | web wheel re-invert (new) |
| `src/recyclerview/hooks/useInvertedWheelFix.ts` | native no-op (new) |
| `src/recyclerview/RecyclerView.tsx` | wire up the hook |

## Reviewers' hat-rack 🎩

- [ ] **Web, inverted list** — scroll up/down keeps rendering items (no blank area), wheel scrolls in the correct direction
- [ ] **Native (iOS/Android), inverted list** — no behavior change; smooth scroll, `renderItem` fires as expected
- [ ] `firstItemOffset` stays `~0` while scrolling an inverted list on web (was growing `1057 → 4009 → 9923` before the fix)

### DIFF
--- a/src/recyclerview/RecyclerView.tsx
+++ b/src/recyclerview/RecyclerView.tsx
@@ -52,6 +52,7 @@ import { getInvertedTransformStyle } from "./utils/getInvertedTransformStyle";
 import { StickyHeaders, StickyHeaderRef } from "./components/StickyHeaders";
 import { ScrollAnchor, ScrollAnchorRef } from "./components/ScrollAnchor";
 import { useRecyclerViewController } from "./hooks/useRecyclerViewController";
+import { useInvertedWheelFix } from "./hooks/useInvertedWheelFix";
 import { RenderTimeTracker } from "./helpers/RenderTimeTracker";
 
 /**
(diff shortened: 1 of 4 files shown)


HuuNguyen312:

PR #2361, opened 2026-07-04
fix(web): stop scrollbar-oscillation relayout loop in LinearLayoutManager

### LINKED ISSUE #2334: [WEB] Scrollbar oscillation cause infinite layout loop
## Description

On **web** with classic (non-overlay) scrollbars — i.e. Windows/Linux Chrome and macOS with "Always show scroll bars", FlashList v2 can enter an infinite layout loop and crash with `Maximum update depth exceeded` because the vertical scrollbar's own appearance/disappearance feeds back into the list's cross-axis (width) measurement.

When list content height sits right at the threshold where it just barely overflows the viewport, the browser shows a vertical scrollbar. The scrollbar consumes ~15–17px of width, which reduces the measured client width FlashList uses as [boundedSize](https://github.com/Shopify/flash-list/blob/20e29c20e440a0a8e0492c303c254dbefee3ad62/src/recyclerview/layout-managers/LinearLayoutManager.ts#L26-L28). The narrower width re-flows items so the content no longer overflows → the scrollbar is removed → the measured width grows back → content overflows again → the scrollbar returns → and so on, forever.

This is web-specific: native platforms use overlay scrollbars that don't consume layout width, so they never oscillate.

## Current behavior

- The list rapidly re-lays-out, the scrollbar visibly flickers, and React throws `Maximum update depth exceeded` (`setState` loop in `updateLayoutParams` → recompute layout → measure → `updateLayoutParams` …), crashing the app.
- Happens whenever content size is near the boundary at which adding/removing the scrollbar flips whether the content overflows.

## Expected behavior

FlashList should not let the scrollbar's own width feedback-loop the layout.

## Reproduction

Make sure to test on web with classic (non-overlay) scrollbar. If you use MacOS fo to Settings -> Appearance -> Windows -> Show scroll bars -> Always.

https://snack.expo.dev/@vikstash/flashlist_loop_crash

## Platform

- [ ] iOS
- [ ] Android
- [x] Web (if applicable)

## Environment

```
Paste output here
```

</details>

**FlashList version:** 
2.3.0

## Additional context

Root cause is in `LinearLayoutManager.ts`, method `updateLayoutParams`. `boundedSize` is taken directly from the measured window size with no detection of scrollbar-induced oscillation, so the scrollbar appearing/disappearing keeps changing the measured width and re-triggering layout.

## Checklist

momomuchu:
On web with classic (non-overlay) scrollbars, borderline-overflow content makes the scrollbar toggle on and off every frame, which loops relayout forever and crashes with "Maximum update depth exceeded".

`updateLayoutParams()` recomputed on any `boundedSize` change with no hysteresis, so a ~17px scrollbar-width delta alternated indefinitely. This adds A/B/A oscillation detection: when the cross-axis size alternates between two values within a scrollbar-width threshold (20px), it locks to the narrower one and treats further sub-threshold changes as noise. A real resize past the threshold unlocks and re-baselines; an axis flip clears the lock.

The test is a pure layout-manager unit test that feeds an oscillating width sequence and asserts convergence instead of unbounded recompute. It fails without the fix (10 unbounded recomputes) and passes with it. Layout-manager and grid/masonry suites green, tsc and lint clean.

Closes #2334

### DIFF
--- a/src/__tests__/LinearLayoutManager.test.ts
+++ b/src/__tests__/LinearLayoutManager.test.ts
@@ -288,4 +288,94 @@ describe("LinearLayoutManager", () => {
       expect(getAllLayouts(manager).length).toBe(3);
     });
   });
+
+  describe("Web scrollbar oscillation (issue #2334)", () => {
+    // On web with classic (non-overlay) scrollbars, the scrollbar's own width
+    // (~15-17px) feeds back into boundedSize: showing the scrollbar narrows the
+    // measured width, which can make content fit and hide the scrollbar, which
+    // widens the measurement again, re-triggering the scrollbar... forever.
+    // updateLayoutParams must detect this A/B/A bounce and stop recomputing.
+    const noScrollbarWidth = 400;
+    const scrollbarWidth = 383; // 400 - 17px classic scrollbar
+
+    it("locks boundedSize and stops recomputing once a scrollbar-width oscillation is detected", () => {
+      const manager = createPopulatedLayoutManager(
+        LayoutManagerType.LINEAR,
+        3,
+        defaultParams,
+        100,
+        100
+      );
+      const recomputeSpy = jest.spyOn(manager, "recomputeLayouts");
+
+      const oscillation = [
+        scrollbarWidth,
+        noScrollbarWidth,
+        scrollbarWidth,
+        noScrollbarWidth,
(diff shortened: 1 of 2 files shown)


momomuchu:
I have signed the CLA!

PR #2373, opened 2026-07-08
fix(viewability): exclude items behind sticky header offset

### LINKED ISSUE #2238: viewableItems not working correctly with an absolute header
## Description

I use an absolutely positioned header above my FlashList. I tried setting the `offset` in `stickyHeaderConfig` or `progressViewOffset` to match the header height, but the `viewableItems` from `onViewableItemsChanged` still include items that are under the header and should not be considered visible.

## Current behavior

`viewableItems` includes items that aren’t actually visible.

## Expected behavior

`viewableItems` should not include items that are under the header.

**FlashList version:** 2.3.1

codergon:
## Description

Fixes #2238

When there is a fixed header on top of the list, you can set `stickyHeaderConfig.offset` so sticky headers stop below it. That offset marks the area covered at the top of the list, but viewability was ignoring it, so items scrolled behind the header were still reported by `onViewableItemsChanged`.

This passes the same offset into the viewability check. The visible area now starts at `offset` instead of 0, so rows hidden behind the header are no longer counted as viewable, and partial visibility is measured against the area below the offset. Nothing changes when no offset is set, since it defaults to 0.

## Reviewers' hat-rack:

- [ ] `ViewabilityHelper.isItemViewable` clamps the top of the visible area to the offset and divides `viewAreaCoveragePercentThreshold` by the height below it
- [ ] `ViewabilityManager` reads `stickyHeaderConfig.offset` and passes it through
- [ ] Behavior is unchanged when no offset is set, covered by the existing tests

List scrolled so a row sits behind a 120px fixed header. Before, that row is still reported viewable. After, it is excluded.

<table>
  <tr>
    <th width="50%">Before</th>
    <th width="50%">After</th>
  </tr>
  <tr>
   <td width="50%" style='border: none'>
        Before: with no offset, rows behind the header are still reported as viewable
    </td>
    <td width="50%" style='border: none'>
        After: with the offset set, rows behind the header are excluded from viewable items
    </td>
  </tr>
</table>

### DIFF
--- a/src/FlashListProps.ts
+++ b/src/FlashListProps.ts
@@ -435,6 +435,9 @@ export interface FlashListProps<TItem>
          * Offset from the top of the list where sticky headers should stick.
          * This is useful when you have a fixed header or navigation bar at the top of your screen
          * and want sticky headers to appear below it instead of at the very top.
+         *
+         * Items hidden behind this offset are also excluded from `onViewableItemsChanged`,
+         * since they aren't actually visible to the user.
          * @default 0
          */
         offset?: number;
--- a/src/__tests__/ViewabilityHelper.test.ts
+++ b/src/__tests__/ViewabilityHelper.test.ts
@@ -251,6 +251,37 @@ describe("ViewabilityHelper", () => {
     );
   });
 
+  it("excludes items hidden behind a top offset (fixed header)", () => {
+    const viewabilityHelper = new ViewabilityHelper(
+      null,
+      viewableIndicesChanged
+    );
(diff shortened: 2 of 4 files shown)


mayvagio2016:
I hope this PR gets merged soon.

mayvagio2016 (approved):

codergon:
I have signed the CLA

codergon:
Hi @naqvitalha  
Could you take a look at this when you get a chance? Small, low-risk fix, already has one approval. Thanks!

Issue #638, opened 2022-10-18
ItemSeparatorComponent with `reverse={true}`: not computed properly


arnaudambro:
## Current behavior

I'm building a chat so the FlashList looks like

```jsx
<FlashList
  data={messages}
  estimatedItemSize={20}
  inverted
  renderItem={ChatMessage}
  ItemSeparatorComponent={props => <Text>{JSON.stringify(props)}</Text>}
/>
```

I expect the `ItemSeparatorComponent` to be located between the `leadingItem` and the `trailingItem`

## Expected behavior

The `ItemSeparatorComponent` is located AFTER the `trailingItem`.
But because of the `inverted` prop, it's not the case

## Platform:

- [X] iOS
- [X] Android

## Environment

1.3.0

fortmarek:
Please, provide full reproduction steps or a sample using Snack.

Makatun:
Any news on this?

phanthh:
Any updates on this?

bryanltobing:
found this issue as well

Makatun:
Please fix

mkhoussid:
Have any of you tried fixing this yourselves and using [`patch-package`](https://www.npmjs.com/package/patch-package) until Shopify eventually rolls our their own fix?

https://github.com/Shopify/flash-list/blob/main/src/FlashList.tsx#L662

```diff
  private getCellContainerChild = (index: number) => {
    return (
      <>
+        {this.props.inverted ? this.separator(index) : null}
        <View
          style={{
            flexDirection:
              this.props.horizontal || this.props.numColumns === 1
                ? "column"
                : "row",
          }}
        >
          {this.rowRendererWithIndex(index, RenderTargetOptions.Cell)}
        </View>
-        {this.separator(index)}
+        {!this.props.inverted ? this.separator(index) : null}
      </>
    );
  };
```

PR #2441, opened 2026-08-14
fix(recycling): place item separator correctly in inverted lists

HoshangDEV:
## Description

In an inverted list, `ItemSeparatorComponent` lands on the wrong side of the item: every separator shifts one slot toward the start, the last two items sit flush against each other, and a stray separator dangles off the start edge.

`inverted` applies a flip transform to both the scroller (`useSecondaryProps`) and each cell (`ViewHolder`), so the two cancel inside the cell but the cells themselves are mirrored — a cell's trailing edge points at index − 1, not index + 1. The fix is placement only: render the separator ahead of the item when inverted. `leadingItem`/`trailingItem` semantics, the last-row suppression in `ViewHolderCollection`, and measured cell size are all unchanged.

Related to #638, which reported this against 1.3.0 and still reproduces on 2.3.2.

- [ ] The double-flip reasoning in `ViewHolder.tsx` is the whole basis of the fix — worth confirming independently.
- [ ] Grid and masonry (`numColumns > 1`) should follow the same argument, but have no dedicated coverage in this PR.

`InvertedTest` on the iOS simulator (iPhone 17, iOS 26.5). The fixture screen was temporarily tweaked to make the effect legible — separator 12px red instead of 1px grey, and 5 items so both ends of the list fit on screen. Those tweaks are not part of this PR.

| Before | After |
|---|---|
|  |  |

Before: Item 4 and Item 3 are flush with no separator between them, and a stray separator sits below Item 0 against the header. After: a separator between all four adjacent pairs, and both outer edges clean.

## Test plan

- [x] `yarn test` (189 passed, 15 suites), `yarn type-check`, `yarn lint`
- [x] New `src/__tests__/InvertedSeparator.test.tsx` asserts separator-vs-item order per cell in both directions, and was confirmed to fail on unmodified `main` for the inverted case
- [x] Verified on iOS simulator — before/after above
- [ ] Android emulator — not run


HoshangDEV:
I have signed the CLA!

PR #2466, opened 2026-08-26
fix(scroll): drop stale autoscroll latch when autoscrollToBottomThreshold is disabled

pgrignaffini:
## Description

`useBoundDetection` arms `pendingAutoscrollToBottom` while `maintainVisibleContentPosition.autoscrollToBottomThreshold` is enabled and the list sits near the bottom. When the caller later disables autoscroll (sets the threshold to `undefined`/negative), `checkBounds` skips the latch block entirely — so the armed latch survives **frozen** instead of clearing. The next `data` change then consumes it in `runAutoScrollToBottomCheck` and fires an un-animated `scrollToEnd`, even though autoscroll is disabled at that moment.

This stomps caller-owned scroll choreography. Concrete case: a chat surface that keeps `autoscrollToBottomThreshold: 0.2` while idle, then sets it to `undefined` before appending the just-sent message so it can run its own post-send scroll animation. The stale latch fires `scrollToEnd({animated: false})` on the message insert, and (on iOS with a keyboard `contentInset`) overshoots past the intended offset — the message teleports instead of animating.

The fix mirrors the consume-time guard added for `scrollToIndex` in #2290: when consuming the latch, re-read the current threshold and drop the latch without scrolling if autoscroll is disabled. Related history: #1736 / #1737 fixed the *arming* side (`checkBounds` no longer latches while disabled) but left the consume side unguarded for latches armed before the disable.

Affected package: `@shopify/flash-list` (v2, `src/recyclerview/hooks/useBoundDetection.ts`).

- [ ] `yarn jest src/__tests__/RecyclerView.test.tsx` — the new test "does not fire scrollToEnd from a latch armed before autoscrollToBottomThreshold was disabled" fails on `main` and passes with this change; the adjacent #2290 tests still pass.
- [ ] Behavior check: with the threshold enabled the existing positive-path test ("fires scrollToEnd from the autoscroll path when isOffsetProjectionEnabled is true") is unaffected — the guard only engages when the threshold is unset/negative at consume time.

Repro shape (chat app): open a list settled at the bottom (latch arms) → disable `autoscrollToBottomThreshold` → append an item → list jumps to the end un-animated. With this fix the append no longer scrolls, leaving scroll position to the caller.


pgrignaffini:
I've signed the CLA.

PR #2472, opened 2026-08-28
fix: forward extraData to the documented getItemType callback

OskarEichler:
## Fix

- pass the current `extraData` as the documented third argument of `getItemType`
- add a focused regression that exercises the real layout-manager call path

## Problem

`FlashListProps` documents `getItemType(item, index, extraData)`, but `RecyclerViewManager` supplied only the first two arguments. Callbacks using the documented third argument therefore received `undefined` during layout and recycling.

## Compatibility / observable changes

No signature change. Callbacks now receive their documented third argument. This PR does not claim that changing only `extraData` reassigns every existing recycled cell; that is a separate behavior and is intentionally excluded.

## Verification

- regression fails on untouched `main` and passes with this fix
- full Jest suite: 14 suites, 188 tests passed
- library TypeScript check passed
- changed-file ESLint and `git diff --check` passed
- no dependency, lockfile, native SDK, or Songstats application change

- [ ] Check the stated compatibility behavior and focused regression.
- [ ] Run the existing test/type/lint commands on a supported environment.

### DIFF
--- a/src/__tests__/RecyclerViewManager.test.ts
+++ b/src/__tests__/RecyclerViewManager.test.ts
@@ -71,4 +71,24 @@ describe("RecyclerViewManager", () => {
       expect(consoleWarnSpy).not.toHaveBeenCalled();
     });
   });
+
+  it("forwards extraData to getItemType", () => {
+    const data = [{ id: 1 }];
+    const extraData = { itemType: "featured" };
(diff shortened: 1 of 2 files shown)


OskarEichler:
I have signed the CLA!

PR #2479, opened 2026-08-28
fix(example): preserve section data and forward contact scroller refs

OskarEichler:
## Fixes

Remove only the selected item in its section, use immutable functional updates and collision-free section/item keys, support adding after the last section is removed and cancel duplicate/pending refreshes. Build static contacts lazily before the first render, memoize sticky indices in one pass and pass ScrollView directly so refs reach the native scroller.

## Compatibility / observable changes

Example-only changes; no library API change. The section removal operation now matches its label. Contact data is available in the first render rather than after a mount effect.

## Verification

- Consumer verification: the combined fixes were packaged on the unchanged npm 2.3.2 runtime baseline and checked with React Native 0.87.1 / React 19.2.3. Immutable install, lint, both Android Debug variants, both unsigned iOS Simulator variants, four Metro bundles, 13 web targets and four browser-extension builds pass. All 298 installed non-metadata package files match the reviewed artifact byte-for-byte.
- Independent patch applied to `435b5141df4960ebbbc0b93264db3f5a4b23fffd`: all 187 existing tests (14 suites), library TypeScript build and changed-file ESLint pass.
- One-off actual-source regression diagnostics pass:
  - sections remove only the tapped item in its section
  - sections can be added after removing every section
  - sections keep distinct header and item keys
  - sections refresh is bounded and cancelled on unmount
  - contacts FlashList has data on its only initial render
  - contacts SectionList has data on its only initial render
  - contacts custom scroller forwards its native ref
  - contacts memoizes sticky indices across context-independent renders
- The combined review workspace passes Android Debug, unsigned iOS Simulator Debug/Release, both production Metro bundles, web production export, Docusaurus/Jekyll builds and all 24 existing iOS Detox tests (10 suites). The contacts comparison was rerun after its final change. These combined-workspace results are not a claim that every device/platform was tested independently for this PR.
- No checked-in tests, reference screenshots, dependency versions or native SDK versions were changed.

- [ ] Check the stated compatibility behavior and reproduce the relevant cases above.
- [ ] Run the existing test/type/lint commands and exercise the affected example or list behavior on your supported device matrix.

Physical-device behavior, Android Detox, assistive-technology conformance and release signing were not verified. No app deployment is part of this PR.

### DIFF
--- a/fixture/react-native/src/SectionList.tsx
+++ b/fixture/react-native/src/SectionList.tsx
@@ -2,7 +2,7 @@
  Use this component inside your React Native Application.
  A scrollable list with different item type
  */
-import React, { useMemo, useRef, useState } from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import {
   View,
   Text,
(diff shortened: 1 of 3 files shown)


OskarEichler:
I have signed the CLA!

Issue #2440, opened 2026-08-13
ViewHolderCollection render throws "index out of bounds, not enough layouts" when the render stack outlives a layout-table shrink


hoangzinh:
## Description

`ViewHolderCollection` can throw `index out of bounds, not enough layouts` while rendering, taking down the tree. It happens when the render stack (`RenderStackManager.keyMap`) still holds an entry whose index was removed from the layout table (`LayoutManager.layouts`): the render path reads layouts through an unguarded `getLayout(index)`, which throws instead of returning `undefined`.

Same error and same `getLayout` family as #2291, different call site — that one is the `validateItemSize` measurement callback, this one is the render path.

## Current behavior

`RecyclerViewManager.modifyChildrenLayout()` truncates the layout table on its first line, but prunes the render stack only on some of its exit paths ([`RecyclerViewManager.ts#L259-L281`](https://github.com/Shopify/flash-list/blob/4c6abc3b8efbcc9b2556dd8aa27c40ffdf875d34/src/recyclerview/RecyclerViewManager.ts#L259-L281)):

```ts
modifyChildrenLayout(layoutInfo, dataLength) {
  this.layoutManager?.modifyLayout(layoutInfo, dataLength);   // truncates layouts when shrinking
  if (dataLength === 0) return false;                          // render stack not pruned
  if (this.layoutManager?.requiresRepaint) { …; return true; } // render stack not pruned
  if (this.hasRenderedProgressively) {
    if (!this.isFirstPaintOnUiComplete) return false;          // render stack not pruned
    return this.recomputeEngagedIndices() !== undefined;       // prunes ONLY if the engaged range changed
  } else {
    this.renderProgressively();                                // this branch does sync the render stack
  }
  …
}
```

[...]

Issue #2291, opened 2026-05-22
Stale ViewHolder onLayout can throw when its layout index no longer exists


KJ21-ENG:
## Description

`ViewHolder.onLayout` can report an item size after the list's data/layout table has already changed. When that happens, `validateItemSize()` still uses the render-time `index` captured by the `ViewHolder`, and currently reads the stored layout with `recyclerViewManager.getLayout(index)`.

If the stale index is no longer present in the layout manager, `getLayout(index)` throws `index out of bounds, not enough layouts`, which can crash the app during fast list transitions or navigation away from a screen.

The relevant current code path is:

- `validateItemSize()` calls `recyclerViewManager.getLayout(index)`:
  https://github.com/Shopify/flash-list/blob/f4278ba0e0aa0da4f8b53035350fc5e063078a54/src/recyclerview/RecyclerView.tsx#L374-L377
- `tryGetLayout()` already exists and safely returns `undefined` when the index is outside the current layout table:
  https://github.com/Shopify/flash-list/blob/f4278ba0e0aa0da4f8b53035350fc5e063078a54/src/recyclerview/RecyclerViewManager.ts#L138-L153
- `LayoutManager.getLayout()` throws when `index >= this.layouts.length`:
  https://github.com/Shopify/flash-list/blob/f4278ba0e0aa0da4f8b53035350fc5e063078a54/src/recyclerview/layout-managers/LayoutManager.ts#L230-L233

## Current behavior

When an obsolete `ViewHolder` layout callback runs after the underlying FlashList layouts have changed, `validateItemSize()` calls:

```ts
const layout = recyclerViewManager.getLayout(index);
```

If that render-time `index` no longer exists, `LayoutManager.getLayout()` throws:

```text
index out of bounds, not enough layouts
```

This can happen during fast data changes or screen transitions where React Native/native layout callbacks from old cells arrive after FlashList has already updated the layout table.

## Expected behavior

Stale layout callbacks for indexes that no longer exist should be ignored. Current indexes should continue through the existing width/height validation.

A minimal defensive change is to use `tryGetLayout(index)` in `validateItemSize()`:

```ts
const layout = recyclerViewManager.tryGetLayout(index);
if (layout === undefined) {
  return;
}
```

That preserves the existing validation for current layouts while avoiding a crash for obsolete measurements.

## Reproduction

**Expo Snack or minimal reproduction link:**

[...]

PR #2507, opened 2026-09-09
fix(sticky-headers): guard binary search against unmeasured layouts

priyanshu-cashbook:
## Description

`StickyHeaders.compute()` throws `index out of bounds, not enough layouts` out of the scroll handler, taking the tree down. Seen in production on 2.0.2 across iOS and Android — 45 users in four days on a single screen, all of them a paginating list with `stickyHeaderIndices`.

### The window

`compute()` bails out on a data-prop length but then reads positions from the layout table:

```ts
const lengthInvalid =
  sortedIndices.length === 0 ||
  recyclerViewManager.getDataLength() <= sortedIndices[sortedIndices.length - 1];
```

- [`getDataLength()`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/RecyclerViewManager.ts#L274) is `props.data.length`, set by `updateProps()` during render.
- [`getLayout()`](https://github.com/Shopify/flash-list/blob/main/src/recyclerview/layout-managers/LayoutManager.ts#L162) throws on `index >= this.layouts.length`, and `layouts` only grows in `modifyLayout()`, reached from `modifyChildrenLayout()` after the children have laid out.

So when a page is appended, `data` and `stickyHeaderIndices` update together in one render and `lengthInvalid` is satisfied for the new trailing header, but `layouts` is still the previous length for at least one more commit. `onScrollHandler` calls `stickyHeaderRef.current?.reportScrollEvent()` → `compute()`, and any scroll event landing in that gap throws.

It needs a list that both paginates and sets `stickyHeaderIndices`, which is why it is not constant. Shrinks are safe (the table is truncated on the first line of `modifyLayout`, so it is never *shorter* than the data); only growth exposes it.

### Fix

Read through `tryGetLayout()` and treat a missing layout as below the viewport.

The two reads immediately below the binary search already do exactly this, and #2460 assumes this file is already safe on that basis (`"as StickyHeaders, the measurement effect and the public getLayout ref already do"`) — the `getY` callback is the one read in here that was still unguarded. That PR covers `ViewHolderCollection` and `validateItemSize` on the *shrink* path, so the two are complementary rather than overlapping; this one needs no coordination with it.

`Number.MAX_SAFE_INTEGER` is also the correct answer and not just a non-throwing one: what goes unmeasured is the freshly appended tail, which really is below everything measured, so the last measured header stays selected instead of one that may not be on screen yet. The guard is inert whenever the index is valid, so no list that is not already in this state changes behaviour.

- [ ] `src/__tests__/StickyHeaders.test.tsx` — two cases added, driving the real component through a manager whose `getLayout` throws past `measuredCount` and whose `tryGetLayout` returns `undefined`, matching `RVLayoutManager` / `RecyclerViewManager`. Both fail on `main` with the exact production message and pass with the fix; the 19 existing cases in the file are untouched by it.
- [ ] Worth a second opinion on the sentinel: bailing out of `compute()` early instead was the other option, but `lengthInvalid` is derived from data rather than layouts, so nothing would re-run `compute()` once the layouts caught up and the header could stay stale until the next scroll event.
- [ ] `yarn test --forceExit` 189 passed / 14 suites, `yarn type-check` and `yarn lint` clean.

Related: #2440, #2291, #2460.


priyanshu-cashbook:
I have signed the CLA

